PR #12, opened 2015-02-25
Diverse bug fixes and patches

mklassen:
This pull request includes a number of patches for diverse issues. I can split the request if that is desired.

1. Added const declaration to methods that are constant.
2. Fixed incorrect getNumberOfDataElements() method.
3. Added missing Image<T>::setFlags() method and also added methods that accept const FlagBit & arguments for setting flags.
4. Changed some by value methods to by const reference and fixed some bugs in get/setAttributeString for Image<T>
5. Added ability to disable int typedefs to allow use when typedefs are already provided, e.g. a previous typedef short int16_t  is incompatible with a second typedef __int16 int16_t.
6. Set a minimum boost version of 1.43 as I was unable to compile with earlier versions.
7. Added some explicit casts to suppress compiler warnings.

### DIFF
--- a/include/ismrmrd/ismrmrd.h
+++ b/include/ismrmrd/ismrmrd.h
@@ -22,12 +22,14 @@
 /* Language and cross platform section for defining types */
 /* integers */
 #ifdef _MSC_VER /* MS compiler */
+#ifndef HAS_INT_TYPE
 typedef __int16 int16_t;
 typedef unsigned __int16 uint16_t;
 typedef __int32 int32_t;
 typedef unsigned __int32 uint32_t;
 typedef __int64 int64_t;
 typedef unsigned __int64 uint64_t;
+#endif
 #else /* non MS C or C++ compiler */
 #include <stdint.h>
 #include <stddef.h>     /* for size_t */
--- a/include/ismrmrd/ismrmrd.h
+++ b/include/ismrmrd/ismrmrd.h
@@ -367,6 +367,7 @@ EXPORTISMRMRD size_t ismrmrd_size_of_ndarray_data(const ISMRMRD_NDArray *arr);
  */
 EXPORTISMRMRD bool ismrmrd_is_flag_set(const uint64_t flags, const uint64_t val);
 EXPORTISMRMRD int ismrmrd_set_flag(uint64_t *flags, const uint64_t val);
+EXPORTISMRMRD int ismrmrd_set_flags(uint64_t *flags, const uint64_t val);
 EXPORTISMRMRD int ismrmrd_clear_flag(uint64_t *flags, const uint64_t val);
 EXPORTISMRMRD int ismrmrd_clear_all_flags(uint64_t *flags);
 /** @} */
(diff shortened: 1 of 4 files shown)


naegelejd:
I have a couple questions:

1. I disagree that the `getDataPtr` and `getTrajPtr` member functions are `const`. They return raw pointers to the underlying data, and changing this data certainly changes the corresponding `Acquisition` or `Image`. Do you disagree?
1. What is the purpose of `ismrmrd_set_flags`? If you want to set the entire `uint64_t flags` field of an `ISMRMRD_AcquisitionHeader` or `ISMRMRD_ImageHeader` why wouldn't you just set the flags? e.g.

    ```c
    ISMRMRD_AcquisitionHeader* head;
    head.flags = ISMRMRD_ACQ_FIRST_IN_ENCODE_STEP1 | ISMRMRD_ACQ_FIRST_IN_SLICE;
    ```
    A fair argument **for** your change would be that there is an `ismrmrd_clear_all_flags` function, which is the same as `head.flags = 0`.
3. Did you intentionally leave the non-`const` versions of `getHead` and `getDataPtr` for `class Image` in 20dacf4 when adding the `const` versions? We should decide on the intended behavior.
 
Otherwise, this looks good.

ghisvail:
> I disagree that the getDataPtr and getTrajPtr member functions are const. They return raw pointers to the underlying data, and changing this data certainly changes the corresponding Acquisition or Image. Do you disagree?

But the changes to the raw data ptr would be made in a different scope, which is irrelevant to const correctness. Adding const is here to signal that calling the method does not change the state of the object, which should be the case for any getter method, no ?

mklassen:
1. ```getDataPtr``` and ```getTrajPtr``` are ```const``` because they return ```const``` pointers, i.e. the caller is not "allowed" to change the underlying data without a ```const_cast```, in which case everything ```const``` is irrelevant and it is up to the programmer to make sure they do not break something critical. There should probably be both ```const``` and non-```const``` versions (see point 3).

2. The ```ismrmrd_set_flags``` is a convience function for the C++ interface which had a ```setFlags()``` method declared, but not implemented. It was to mirror the same structure of all the other flag setting methods in the C++ interface.

3. Yes, it was intentional to leave both the ```const``` and the non-```const``` version of ```getDataPtr```. This allows you to get a const pointer from a ```const Image<T>```, and also non-```const``` or ```const``` pointer if you have an ```Image<T>```. If you do not have the ```const``` one it is impossible to get a pointer from a ```const Image<T>``` without resorting to a ```const_cast```. If you do not have the non-```const``` one it is impossible to get a pointer to data that you can change without resorting to a ```const_cast```. Since users should not be ```const_cast```'ing, you need both.

inati:
In my opinion, the const label is completely useless from a programming safety point of view.  If you ask for a pointer to the raw data block and I give it you, then the contract is that you are on your own and operating without a safety net and you had better know what you are doing.  You are free to const_cast away if you like and there is nothing I can do to stop you.

inati:
I had left out the setFlags in both Acquisition and Image to encourage people to set the flags one at a time, but maybe that's an unnecessary amount of handholding.  If that's the case, then we should take the const of the Acquisiton::flags() function.

naegelejd:
1. OK. Ghislain is correct in that what a user does with the pointer obtained from `getDataPtr` is irrelevant to `const`-correctness. In this case there should probably be both `const` and non-`const` versions.
2. This makes sense. I question the utility of both `ismrmrd_set_flags` and `ismrmrd_clear_all_flags`, but that's a different issue.
3. Great, this also makes sense. However, I expected `class Acquisition` and `class NDArray` to also have `const` and non-`const` versions of similar member functions.

mklassen:
The non-```const``` versions of ```getDataPtr``` and ```getTrajPtr``` have been added in commit  48a2776

PR #18, opened 2015-03-13
Unit tests and Travis-CI configuration

naegelejd:


naegelejd:
This fails to compile on Windows because I'm only checking if `Boost_FOUND` is true in the `tests/` directory. I'll change that to `Boost_UNIT_TEST_FRAMEWORK_FOUND` and cross my fingers.

inati:
Don't forget to sacrifice a chicken.

PR #22, opened 2015-04-03
making HDF5 optional

naegelejd:


ghisvail:
I'll try it for Windows on my Philips VM (VS 2008). It won't happen before Wednesday.

naegelejd:
Updated to address some issues @inati mentioned:

- CMake now issues a warning if HDF5 is not found
- `ismrmrd_info` now prints whether ISMRMRD is installed with dataset support

inati:
Update minor version number?

naegelejd:
If you're asking me I presume you mean "increment the patch number". Major is for breaking changes to fixed memory layout structs, minor is for changes to XML schema.

PR #24, opened 2015-04-23
For upstream

mklassen:
Added ```const char *``` accessors for ```Image<T>``` attribute string to simplify copying to/from ```char *``` by avoiding creation of intermediate ```std::string``` copy.

Added missing Dataset::readImage instantiations and Boost include for tests.

### DIFF
--- a/include/ismrmrd/ismrmrd.h
+++ b/include/ismrmrd/ismrmrd.h
@@ -755,8 +755,10 @@ public:
     void setHead(const ImageHeader& head);
     
     // Attribute string
-    void getAttributeString(std::string &atrr) const;
+    void getAttributeString(std::string &attr) const;
+    const char *getAttributeString() const;
     void setAttributeString(const std::string &attr);
+    void setAttributeString(const char *attr);
     size_t getAttributeStringLength() const;
     
     // Data
--- a/libsrc/ismrmrd.c
+++ b/libsrc/ismrmrd.c
@@ -268,10 +268,13 @@ int ismrmrd_make_consistent_image(ISMRMRD_Image *im) {
    
     attr_size = ismrmrd_size_of_image_attribute_string(im);
     if (attr_size > 0) {
-        im->attribute_string = (char *)realloc(im->attribute_string, attr_size);
+        // Allocate space plus a null-terminating character
+        im->attribute_string = (char *)realloc(im->attribute_string, attr_size+sizeof(*im->attribute_string));
         if (im->attribute_string == NULL) {
             return ISMRMRD_PUSH_ERR(ISMRMRD_MEMORYERROR, "Failed to realloc image attribute string");
         }
+        // Set the null terminating character
+        im->attribute_string[im->head.attribute_string_len] = '\0';
     }
         
     data_size = ismrmrd_size_of_image_data(im);
--- a/libsrc/ismrmrd.cpp
+++ b/libsrc/ismrmrd.cpp
@@ -943,16 +943,20 @@ template <typename T> void Image<T>::setAttributeString(const std::string &attr)
 
 template <typename T> void Image<T>::setAttributeString(const char *attr)
 {
-    size_t length = strlen(attr);
-    im.head.attribute_string_len = static_cast<uint32_t>(length);
+    // Get the length and include the null terminating character
+    size_t length = strlen(attr) + 1;
 
-    // Add null terminating character
-    length++;
-
-    im.attribute_string = (char *)realloc(im.attribute_string, length);
-    if (NULL==im.attribute_string) {
+    // Allocate space and check for success
+    char *newPointer = (char *)realloc(im.attribute_string, length);
+    if (NULL==newPointer) {
         throw std::runtime_error(build_exception_string());
     }
+
+    // Make changes only if reallocation was successful
+    im.attribute_string = newPointer;
+    im.head.attribute_string_len = static_cast<uint32_t>(length);
+
+    // Copy the string
     strncpy(im.attribute_string, attr, length);
 }
 
--- a/libsrc/ismrmrd.cpp
+++ b/libsrc/ismrmrd.cpp
@@ -931,9 +931,19 @@ template <typename T> void Image<T>::getAttributeString(std::string &attr) const
       attr.assign("");
 }
 
+template <typename T> const char *Image<T>::getAttributeString() const
+{
+   return im.attribute_string;
+}
+
 template <typename T> void Image<T>::setAttributeString(const std::string &attr)
 {
-    size_t length = attr.length();
+    this->setAttributeString(attr.c_str());
+}
+
+template <typename T> void Image<T>::setAttributeString(const char *attr)
+{
+    size_t length = strlen(attr);
     im.head.attribute_string_len = static_cast<uint32_t>(length);
 
     // Add null terminating character
@@ -943,7 +953,7 @@ template <typename T> void Image<T>::setAttributeString(const std::string &attr)
     if (NULL==im.attribute_string) {
         throw std::runtime_error(build_exception_string());
     }
-    strncpy(im.attribute_string, attr.c_str(), length);
+    strncpy(im.attribute_string, attr, length);
 }
 
 template <typename T> size_t Image<T>::getAttributeStringLength() const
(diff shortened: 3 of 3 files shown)


ghisvail:
Following your logic, specialized instantiations of readNDArray for each data type are also missing...

mklassen:
Added the missing readNDArray instantiations as well

naegelejd:
Looks good to me.

inati:
The char* are really not safe.  There is no guarantee that they are null terminated.

inati:
@naegelejd  This is a style choice here, but I remember having a discussion about not returning a pointer to the attribute string (in the C++ api) because it wasn't belt and suspenders safe and because you need to keep track of the string length in the header, etc.  I fine with relaxing that if you think it's okay.

inati:
@mklassen, I think I'm missing something.  Why do we need specific installations of a templated function like readImage?

inati:
oh good lord that was dumb.  sorry, i thought i was in the header file and not the cpp file.  ignore the comment about the templated functions.

naegelejd:
Ahh yes good point @inati. Except that means [this](https://github.com/ismrmrd/ismrmrd/blob/ab5b0d215b3ac7069437e07ccd6f81c7740bdc57/libsrc/ismrmrd.cpp#L929) isn't currently guaranteed to work. `std::basic_string.assign(const char*)` relies on the existence of a nul-terminator. If we "support" non-nul-terminated attribute strings, then we can fix this by changing it to `attr.assign(im.attribute_string, im.head.attribute_string_len)`.

inati:
@naegelejd, I'm conflicted about this.  At the moment, our spec is just char* with attribute length, i.e. it's just a block of memory, and we do nothing to insure that the there is a null terminator i.e. that is a proper c-string.  (Of marginal relevance, in the hdf5 file, AFAIK VLEN strings are not null terminated).  Ideally, we should support non-null terminated strings all the way and fix whatever hanging bugs like the one you mentioned.

ghisvail:
What is attribute_string supposed to contain eventually ?

inati:
This is not really specified.  It's up to the application.

mklassen:
The latest change switches completely away from null terminated strings and therefore obviously requires you to specify length with ```setAttributeString(const char *attr, size_t length)```. It uses ```memcpy``` instead of ```strncpy``` which could have truncated a ```std::string``` that contained non terminating null characters.

inati:
I may have sent us all down a rabbit hole.  What we are talking about is fine for an in-memory api .  BUT, I am worried that we are not going to be able to put a string with a null in the middle of it into an HDF5 file.  Allowing nulls inside the attribute string will lead to the possibility of weird behavior on write.  Or other interfaces (e.g. python via h5py) will not be able to accommodate them.  Or other languages (C++11?) might not allow strings with nulls.

Could we pause for a minute and look at this carefully?  This is one of those weird little big decisions where we could break things without meaning too.  We called it attribute_string for a reason.  Chime in with ideas here please and let's discuss with @hansenms at tomorrow's t-con.

inati:
@mklassen, perhaps you want to split out the read instantiations and the boost test thing into a different PR so that we can move on those while we think more about this string business.

naegelejd:
So why do we have an `attribute_string_len` field in the `Image` header again? Is it something we added out of necessity? Why isn't the attribute string just defined to be a null-terminated sequences of chars (hence the *string* part of its name)?

HDF5 is the **only** thing I can think of that has the limitation of not being able to store arbitrary binary data in variable length strings, which is unfortunate.

ghisvail:
Is that the reason why the ISMRMRD_Image type attributes (head, attribute_string and data) are stored as separate datasets (named header, attributes and data) in the HDF5 file, as opposed to one single entry like for the ISMRMRD_Acquisition type ?

mklassen:
I have created a another PR without the string issues.

naegelejd:
@mklassen, I merged the other PR.
@ghisvail, I believe @inati has a couple reasons why he decided to store Images differently (I think related to HDF5 design).
@inati, we discussed the `attribute_string` this morning and agree that it should be specified as a null-terminated sequence of chars since it is called a "string".
@mklassen, sorry about this, but would you mind reverting your changes to just accept a null-terminated `const char*` and I can merge then?

mklassen:
@naegelejd I will revert the changes. Is attribute_string_len defined to include or exclude the null terminating character?

mklassen:
Based on the C api implementation ```attribute_string_len``` must include the null terminating, so I have made that change as well.

naegelejd:
Unfortunately I think this also need to be discussed. @inati, @hansenms, @ghisvail, @dchansen: what are your thoughts?

The null-terminator is a C/C++ implementation detail (and in some sense, a convention). I think, in terms of the ISMRMRD spec, `attribute_string_len` should represent the length of the actual `attribute_string` content. This makes sense in most other languages, even in C++: `std::string.size()` is the length of the string content, whether it's null-terminated or not. Also, when "streaming" an Image, the null-terminator does not **need** to be included in the `attribute_string`, and `attribute_string_len` essentially exists for streaming purposes, e.g. on a socket.

I know the current C implementation counts the null-terminator, but I think this should change as well.

inati:
I knew this was going to be a mess.  I think @naegelejd is correct that the length in the header should not include the null and that the null is just a c implementation detail so that we can treat it as a c-string and that it should not be in the hdf5 or on the wire. But I want to hear what @hansenms has to say before I fully commit.

hansenms:
I agree, the length of the string should not include the null terminator.

ghisvail:
Just to be clear, say I am implementing my own client and want to de-serialize from bytes to ISMRMRD_Image. Then I am supposed to actually read the attribute_string_len from the fixed size header, add 1 to it and cast / construct from the next set of bytes to a string-like object ? 

What about if the string attribute is actually empty ? Do I still read 1 byte containing just a null terminator ?

Thanks for clarifying.

inati:
The format does specify a particular serialization.  This would depend on your application.

naegelejd:
You should **not** send or receive a null-terminator though. Pretend you're sending your ISMRMD Image to a Python application over TCP/IP... you do not want a null-terminator in the attribute string because they aren't terminators in python, they are just zero bytes and perfectly valid in strings.

The `attribute_string_len` says how many non-zero bytes are in the attribute string, which is the number of bytes that should be sent/received when streaming if you want to avoid problems.

ghisvail:
You guys still have not given me a clear answer. Assuming I read an ISMRMRD_ImageHeader from a stream of bytes received from Gagdetron, with `attribute_string_len=128`, do I read 128 bytes or 129 bytes afterwards to unpack into `attribute_string` ? What happens if `attribute_string_len=0`, do I just unpack the bytes into `data` straight after or is there a dummy `\0` in between ?

hansenms:
You read 128 bytes. The nulls don't go on the wire.

You can read the code in the Gadgetron to see what it does.

------

naegelejd:
I agree with Michael. You read 128 bytes because the null-terminator is C-specific.

That being said, the Gadgetron is actually not a good place to look right now. As far as I know, the only code sending/receiving ISMRMRD Images is in the Gadgetron at the end of a recon chain (MRIImageWriter) and by the `gadgetron_ismrmrd_client` (GadgetronClientImageMessageReader). The current implementation actually **does** send/receive the null-terminator. I think this should be fixed (relevant line [here](https://github.com/gadgetron/gadgetron/blob/d6a91af28668c82ffb80e2099fe166fa36b026f9/gadgets/mri_core/MRIImageWriter.h#L69))

ghisvail:
Same for `send_gadgetron_parameters` and `send_gadgetron_configuration` in `gadgetron_ismrmrd_client` if I am not mistaken.

hansenms:
That should be fixed, but either way ismrmrd does not specify an on the wire protocol, so this seems like something that is a Gadgetron discussion and/or bug. 

------

inati:
I have opened a new issue about the null terminator on the gadgetron project.  Where do we stand with this pull request?  Do we always want to allocate memory for the attribute string and put a single NULL in it if the length is zero?  Should we expose it as a "regular" c-string?

naegelejd:
I think it's fine the way it is. If the attribute string is empty, it has a length of zero and the underlying `char*` is still `NULL`.

I'm merging these changes from @mklassen since they only add the ability to get/set a `char*` attribute string and don't change any existing logic.

PR #26, opened 2015-05-06
add sequence type to ismrmrd sequence parameter section

gtdeveloper5:
Add an optional field "sequence_type" in the sequenceParameters section.


gtdeveloper5:
The ismrmrd is updated for the sequence parameter section, as discussed.

PR #28, opened 2015-06-25
Add ability to read HDF5 files in read-only mode

isolovey-robarts:
Currently one must have write permissions to the ismrmrd (HDF5) file in order to read it in. This change adds a fallback read-only call to H5Fopen if the read-write mode call fails.


inati:
We should probably have a state so that we can give a meaningful error when the user tries to perform a write operation.

hansenms:
@inati, fair point, but isn't this PR better than what we have after all? @naegelejd, thoughts?

I propose we merge this and then we can add the state later??

naegelejd:
Looks good

inati:
Okay.   @naegelejd let's discuss the state and/or a way to indicate a warning.

PR #46, opened 2015-11-18
Patches

mklassen:
1. Modified `realloc()` usage to maintain valid pointers on failure. This allows a calling functions the opportunity to attempt recovery from the failure without leaking memory.
2. Modified the flag manipulation functions to explicitly using `uint64_t` type for flag masks, otherwise the mask might be a smaller type and therefore incorrect. 
3. Modified the dataset HDF5 reading and writing of header and attribute_string to use variable length `char` arrays instead of null-terminated variable length string. Attribute string handling functions are also modified to support `char` arrays which may contain null characters. This requires specifying the array pointer and length to manipulate attribute strings. This change is motivated by the fact that attribute_string and headers are the two locations which might contain confidential patient data and therefore are candidates for encryption and most encryption algorithm will have null characters within the ciphertext and therefore cannot be treated as simple null terminated strings. Libraries compiled with these changes were able to read datasets written with previous versions. The C-style null-terminated string interfaces are maintained to support backwards compatibility with existing software.

### DIFF
--- a/libsrc/ismrmrd.c
+++ b/libsrc/ismrmrd.c
@@ -133,20 +133,22 @@ int ismrmrd_make_consistent_acquisition(ISMRMRD_Acquisition *acq) {
     
     traj_size = ismrmrd_size_of_acquisition_traj(acq);
     if (traj_size > 0) {
-        acq->traj = (float *)realloc(acq->traj, traj_size);
-        if (acq->traj == NULL) {
+        float *newPtr = (float *)realloc(acq->traj, traj_size);
+        if (newPtr == NULL) {
             return ISMRMRD_PUSH_ERR(ISMRMRD_MEMORYERROR,
                           "Failed to realloc acquisition trajectory array");
         }
+        acq->traj = newPtr;
     }
         
     data_size = ismrmrd_size_of_acquisition_data(acq);
     if (data_size > 0) {
-        acq->data = (complex_float_t *)realloc(acq->data, data_size);
-        if (acq->data == NULL) {
+        complex_float_t *newPtr = (complex_float_t *)realloc(acq->data, data_size);
+        if (newPtr == NULL) {
             return ISMRMRD_PUSH_ERR(ISMRMRD_MEMORYERROR,
                           "Failed to realloc acquisition data array");
         }
+        acq->data = newPtr;
     }
 
     return ISMRMRD_NOERROR;
@@ -269,20 +271,22 @@ int ismrmrd_make_consistent_image(ISMRMRD_Image *im) {
     attr_size = ismrmrd_size_of_image_attribute_string(im);
     if (attr_size > 0) {
         // Allocate space plus a null-terminating character
-        im->attribute_string = (char *)realloc(im->attribute_string, attr_size + sizeof(*im->attribute_string));
-        if (im->attribute_string == NULL) {
+        char *newPtr = (char *)realloc(im->attribute_string, attr_size+sizeof(*im->attribute_string));
+        if (newPtr == NULL) {
             return ISMRMRD_PUSH_ERR(ISMRMRD_MEMORYERROR, "Failed to realloc image attribute string");
         }
+        im->attribute_string = newPtr;
         // Set the null terminating character
         im->attribute_string[im->head.attribute_string_len] = '\0';
     }
         
     data_size = ismrmrd_size_of_image_data(im);
     if (data_size > 0) {
-        im->data = realloc(im->data, data_size);
-        if (im->data == NULL) {
+        void *newPtr = realloc(im->data, data_size);
+        if (newPtr == NULL) {
             return ISMRMRD_PUSH_ERR(ISMRMRD_MEMORYERROR, "Failed to realloc image data array");
         }
+        im->data = newPtr;
     }
     return ISMRMRD_NOERROR;
 }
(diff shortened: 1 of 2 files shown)


inati:
changing to a variable length string is a major change.  It may break compatibility with existing files. The reason we are not using variable length strings is that older versions of Matlab did not support it properly.  All this may have changed but testing is required.

hansenms:
This is going to require some discussion. 

As I recall it, there are some issues with how these variable length strings are handled in different environments (e.g. Matlab) and some of the choices were based on compatibility under those conditions. 

We also have to have a very careful look at the memory management yet. There are some low level changes here that I cannot merge without careful study and I don't have time right now. 

@naegelejd or @inati do you have time to look at this.

hansenms:
Just looking at this briefly:

1. The changes to realloc complicate things a bit and I am not sure I understand the use case, what kind of situation do you imagine where such a "recovery" would be meaningful? Do you have a real-world use use case?
2. Sure, but as of yet, I have not seen problems with the current code. Do you have a test case that demonstrates the problem?
3. I think not. The ripple effects of these changes could be pretty dramatic and the proposed use case of storing these encrypted is not supported nor would it be implemented that way. 

Let's hear what the others say, but these are pretty drastic low level changes and I am not sure I have seen an actual problem that they solve. 

@mklassen could you provide some specific use cases and test so that we might understand what you are trying to do here. Have you tested the files that this produces in Matlab, Python, etc.

naegelejd:
My thoughts:

1. This is perfectly reasonable, but a moot point if we plan to merge my 2.x changes that revert the ISMRMRD API to pure C++. See [here](https://github.com/naegelejd/ismrmrd/tree/2.x)
2. This is technically correct. I prefer the literal syntax `1ULL`. See sample program and output below.
3. We started to address the API portion of this in April in #24. I have no other input right now.

Regarding flags, literals and integer widths:
```c
#include <stdio.h>
#include <stdint.h>

#define ARRAY_LENGTH(array) (sizeof(array)) / sizeof((array)[0])

int main(void)
{
    uint64_t chan = 42;

    uint64_t masks[] = {
        1ULL << (chan % 64),
        1LL << (chan % 64),

        1U << (chan % 64),
        1 << (chan % 64),

        (unsigned long long)1 << (chan % 64),
        (signed long long)1 << (chan % 64),
        (unsigned long)1 << (chan % 64),
        (signed long)1 << (chan % 64),

        (unsigned int)(1) << (chan % 64),
        (signed int)(1) << (chan % 64),
        (unsigned short)(1) << (chan % 64),
        (signed short)(1) << (chan % 64)
    };

    for (int i = 0; i < ARRAY_LENGTH(masks); ++i) {
        printf("%llu\n", masks[i]);
    }

    return 0;
}
```

Output on my machine:
```
4398046511104
4398046511104
1024
1024
4398046511104
4398046511104
4398046511104
4398046511104
1024
1024
1024
1024
```

mklassen:
1. `alloc` failures are often due to memory exhaustion, the caller may be able to free additional memory and try again. 
2. @naegelejd demonstrated the issue. I can change the commit to use `ULL`. I intentionally choose not to use the string literal because the flag is of type `uint64_t` which does not have to be `unsigned long long`. I thought the explicit cast assisted the reader in making clear that the constant needed to be the same size as the flag. Since `ULL` is at least 64 bits, it would be safe to change to a string literal in this case.
3. I have dropped this commit as it appears to have the most issues. I can read the generated files in matlab, but you do have to convert  from numeric array of `char` to string interpretation using `char(array)`, so it would require changes to the matlab API. Encryption is my use case for these changes. Although there are prototype encryption filters for HDF5, they do not support encryption of variable length data types. Currently the only way to store encrypted variable length data within HDF5 is to encrypt it separately.

hansenms:
Sorry for my quick replies earlier. Should have taken more time to explain. 

1. I do get how failed allocs occur, my comment was more along the lines of a typical application that would bail out from that in any other way than exiting, but there are actually multiple such scenarios and it is good to support that. 
2. Again, I understand what the fundamental problem is, I was talking about a use case where it occurs when using the ISMRMRD library. Again, such a situation can occur, so this is fine. 
3. Yes, we should look at this separately. I am a bit confused that you want to change to a variable length field for the header if that does not support encryption but the goal was encryption, but that is a detail. The specification calls for an xml file in the header and not an encrypted document, so we need to think about how this would be supported in general. As far as the null termination goes, you could simply base64 encode the encrypted information. I would actually recommend that you have most of the header in clear text and put a base64 encoded (and encrypted) field in the user fields. Then you can stick the patient information in there. Since your files would not be useable by anybody else in their current form, that solution would be better. That way the images could still be reconstructed but only people with the key can decode the patient information. In the long term we could add an encrypted patient information section, but it should be base64 encoded so that it can have a meaningful representation as a string. @mklassen, thoughts?

In any event, I am merging this now.

ghisvail:
Just to let you guys know that the explicit cast to `uint64_t` solved most test failures when building ISMRMRD on `arm` and `powerpc` architectures in [Debian](https://buildd.debian.org/status/package.php?p=ismrmrd). Thanks @mklassen for providing the fix.

mklassen:
@hansenms base64 encoding is very useful concept. It gets around my issues with cipher texts containing null characters and makes my proposed changes unnecessary, thanks.
I can see the selective encryption concept being useful in specific cases, but believe it to be deficient for general data protect because
1. The data can still be reconstructed. Since there are some MR images that permit identifying the patient with just the reconstructed image, a general data protection scheme needs to prevent reconstruction. I actually opt for encrypting the data as well as the xml headers and attribute strings.
2. It presents continuing maintenance issues. Because it is not possible to know with 100% certainty what fields might be security related in every end users environment, this choice should be left to the end user. This means that with every software update the end user has to review any changes to the xml header content and update their encryption policy. Complete encryption of the xml header removes this maintenance issue and minimizes potential security issues.

hansenms:
@mklassen I understand your concerns, but I think it is stepping a bit outside what the purpose of the data format is and what current vendor formats provide. Non of them are encrypted. And one step further, DICOM images (which are a much bigger problem) are not encrypted. The industry standard for dealing with that is to store them in encrypted storage vehicles such as encrypted folders or encrypted hard drives. I believe that would satisfy your requirements and there are much better solutions out there than we could ever provide in the ISMRMRD format. 

Isn't that the way to go for you? Just stick the ISMRMRD files in an encrypted folder on your disk? Are there any downsides to that?

PR #49, opened 2015-11-29
Made first sweep on major changes to data types, etc.

hansenms:
This is for review by @naegelejd 

I have started making some changes to datatypes, etc. Could you have a quick browse and tell me what you think. No need to merge yet, just comment. Will be adding the hdf5 datatypes for images etc. and serialization/deserialization stuff next.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -16,9 +16,9 @@ Questions
 Next steps
 ----------
 
-* Template Acquisition handling in dataset to allow multiple storage types
 * Add Entity interface (serialization/deserialization).
 * Add HDF5 code for images
 * Add appendEntity and readEntity functions.
 * Add classes for other entity types.
 * Add HDF5 datatypes for all entity types.
+* Add consistency checks for entity types and storage types and update tests accordingly.
--- a/TODO.md
+++ b/TODO.md
@@ -1,9 +1,24 @@
 To Do List for 2.x
 ==================
 
+Questions
+----------
+
 * Should the stream number be and argument for append, since it is part of the header?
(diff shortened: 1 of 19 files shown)


naegelejd:
Looks good to me, ignoring the formatting changes.

I don't really like that the storage types are enumerated `SHORT`, `INT`, `INT64`. Since they don't have to mimic C type names, I think it should be either a) `SHORT`, `INT`, `LONG` or b) `INT16`, `INT32`, `INT64`.

I agree that the stream number may be unnecessary in the dataset `append` functions since it's in the entity header.

I prefer your idea of using an integer for dwell time with the explicit `ns` in the name.

I assume there's a need to change all the 16-bit integers in the headers to 32-bit integers.

hansenms:
Thanks @naegelejd , I will change the enumerations. 

On the change from 16-bit to 32-bit. It is not strictly necessary, but when the paper was being revised there was concern about the restrictions of the 16-bit counter indices. So for instance kspace_encode_step_1, if the max number here is 65k, that would seem like plenty, but people might use that index for profile number in a golden angle radial acquisition, in which case it would only really be on the order of a few hundred frames, which could pose a problem. So it made sense to change some of them and then I thought we should just change them all to reduce the chances of confusion and make it a bit more future proof. What do you think?

naegelejd:
I can't think of any downside other than the increased total size of the header, but it's only a 20% increase to the Acquisition/Image header size and it sounds necessary.

hansenms:
OK, I will make some changes and merge this into the 2.x branch when I have time. I will keep you posted as a move through this.

PR #55, opened 2016-06-29
Add cmake package configuration

hjmjohnson:
Provide cleanup and modernization to cmake options to facilitate integration with other packages (i.e. gadgetron).


hansenms:
@hjmjohnson thanks for getting involved and submitting this PR. 

It is clear that you probably have better command of CMake than I do, so your input is much appreciate, but few issues and questions:

1. Did you notice that the travis-ci failed? I have not investigated in detail, but something about the add_dependencies(doc htmldoc) statement. Could you investigate.
2. Also you are going for elimination of a FindIsrmrmrd.cmake script in the Gadgetron, which is fine, but I am concerned about the need to specify ISMRMRD_DIR when configuring downstream projects. Is this a requirement or will defaults be searched?
3. Did you check this on Windows, Linux, and Mac?

I would recommend that you join us for our weekly hangout, which are on Fridays 9am east coast time. I typically post a link to the hangout here: https://groups.google.com/forum/#!forum/gadgetron a couple of minutes before we start. It would be good to talk through these changes before merging. Would you be able to join?

hjmjohnson:
@mjacob75 -- Keeping you in the loop.

@hansenms Just a quick background of my skills:  I am the president of the Insight Software Consoritium and one of the lead developers of the ITK platform.  I have worked with cmake at a moderate level for more than a decade, and recelntly (see the commits for the vxl package on github) I have been expanding my prowess with cmake to a deeper level.

1.  The changes I made yesterday were during the flights back from an NIH review session, and I had very limited internet access.   I did notice the failures but was exhausted when I returned home last night.   I am going to investigate those this morning.

2.  There are several ways to address this.  I would need a little clarification about what you mean by "will  defaults be searched".  There is a cmake variable called "CMAKE_PREFIX_PATH" that will allow for these so called package CONFIG (i.e. ISMRMRDConfig.cmake file in the installed location) files to be found automatically.  If that is an environmental variable, cmake will respect it, and additoinally inside of gadgetron's main CMakeLists.txt file we could append other default paths to optionally set environmental variable.  I am confident that the desired behavior can be achieved.

3. I did build on Mac at the airport, and started a build on Linux.  I did not have access to a windows box from the airport so that was not tested.  Hopefully I can squeeze that in today.

hansenms:
@hjmjohnson thanks for the clarification and introduction. Any contributions that you are able to make would be very much appreciated. The ISMRMRD and the Gadgetron projects have many enthusiastic users and developers, but not all of them make software for a living and that can be seen in places, so style, consistency, robustness improvements would help a lot. 

So you are are saying that if I install ISMRMRD in say /usr/local on Linux and I then run cmake for say the Gadgetron project, I should not need to specify where it is installed (with ISMRMRD_DIR) since it is already a default search path. Or if I have another default install path I should be able to say:

    cmake -DCMAKE_INSTALL_PREFIX=../install  ../

And if ISMRMRD is installed in `../install`, it should be found? I am assuming that is the case, but I wanted to make sure before we merge any of this. 

We have Windows, Linux, and Mac build slaves, so we will know soon enough if it breaks, but since this will be a two-project (ISMRMRD and Gadgetron) operation, it is a bit more of a delicate task and any testing that you can do up front would be great.

hjmjohnson:
@hansenms You are correct.   I can add 1 line to gadgetron main CMakeLists.txt so that /usr/local is always automatically searched.

Here is the psuedo-code for the line

    list(APPEND CMAKE_INSTALL_PREFIX "${CMAKE_INSTALL_PREFIX}:/usr/local")

hjmjohnson:
@hansenms @mjacob75   I will not be able to make tomorrow's 8am CST google hangout meeting.  I have other meetings that can not be moved at that time.

hansenms:
@hjmjohnson no problem about the meeting, we have them every week. Just wanted to make sure that you know about them. We should be able to sort this one out via email, etc.

hjmjohnson:
@mjacob75, @hansenms 

I will very likely not be able to make the google hangout tomorrow because my wifes medical clinic is moving tomorrow, and I am in charge of coordinating all the IT equipment moves.  The IT contractor is supposed to be there at 8am CST (9am EST).

Hans

hjmjohnson:
@hansenms I commented out all the currently unnecessary cmake 2.8.9 features, and now this will compile.  We may eventually want to use these features if we move beyond trivial shared library usage.

hansenms:
OK, I will proceed to merge this later today. We have had a couple of build slaves offline and they just came back and are working their way through some builds. Once I am sure the head of the branches pass all builds, I will merge unless you have concerns?

hjmjohnson:
No more concerns.

 

I just updated gadgetron based on our conversion to automatically find ISMRMRD if it is in a default place, or has a default environmental variable set.

 

Hans

 

 

--

hansenms:
@hjmjohnson after merging this, we actually have a problem. 

In downstream projects, if ISMRMRD is not installed in a standard location, linking doesn't work. Effectively, the `ISMRMRD_LIBRARIES` variable just resolves to `-lismrmrd` when linking and it can't find the library. 

I am struggling with this for our buildslaves right now and not quite sure how to fix it. What you added to the README file about using `-DISMRMRD_DIR:PATH=<>` does not seem to do the trick, e.g.:

    cmake -DISMRMRD_DIR:PATH=../local -DCMAKE_INSTALL_PREFIX=../local ../siemens_to_ismrmrd-source

I am struggling with how to resolve this without having a FindIsmrmrd.cmake file like we did before. 

Do you have any ides/thoughts/comments/suggestions?

hansenms:
@hjmjohnson looks like I needed a `link_directories( ${ISMRMRD_LIB_DIR} )` in the downstream project, which I suppose makes sense. But it does have some downstream effects when people install in different locations, so some projects may be waking up this morning to a bit of pain. But all easily fixed.

hjmjohnson:
I don't have internet right now,   But it appears that the downstream packages must have been depending on full path to the libraries.   Typically I've used link_directories to ISMRMRD_DIRS some where at the top.

hjmjohnson:
:).  Should have read all messages.   We could make it backward compatible easy enough,   But that would not follow the typical usage.   Your call.

hansenms:
I think it is fine. It is just a learning experience for me. 

We will have a problem (unrelated) on our Windows build slave, so once it is back up, I have to make sure it is all working there. 

When you have time, will you update the Gadgetron PR?

PR #56, opened 2016-06-30
Test updated c make version

hjmjohnson:
DO NOT MERGE.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,11 @@
+addons:
+  apt:
+    sources:
+      - george-edison55-precise-backports # cmake 3.2.3 / doxygen 1.8.3
+    packages:
+      - cmake
+      - cmake-data
+ 
 language: cpp
 
 compiler:


hjmjohnson:
DO NOT MERGE:

@hansenms @mjacob75 FYI All the previous build failures were due to travis using cmake version 2.8.7.  

There are some  problems with shared libraries before cmake 2.8.9.  Both the Slicer3D community and the ITK  community, and the VXL communities have a minimum requirement of 2.8.9 to assist with overcomming these problems.  You likely still have a simple enough eco system that you don't have to worry about it yet.

I followed instructions on: https://github.com/travis-ci/travis-ci/issues/4631 to get a newer version of cmake.

This is something for a longer term goal to address.  I made this patch set mostly to document what I found while it is still fresh in my head.

Hans

hansenms:
Thanks for looking into this. Not sure if we should adopt this approach to getting a newer CMake, I am OK with it or should we simply wait until travis updates?

hjmjohnson:
I think we should *NOT* do this right now.  Let's put it on an agenda item for one of the friday hangouts.  I can give the pro's and con's in person better.

If you want to move to better C++11 support, then we will want to jump all the way to at least cmake 3.2 or better.  There are some very nice compiler feature detection capabilities in cmake 3.2 and greater.


Hans

hansenms:
@hjmjohnson I would like to close this if we have no plans to merge. How do you feel about that?

hjmjohnson:
@hansenms   I understand.  I believe that these are important changes, and that they make the building and testing more robust.  I had to pivot away from this effort during the fall, and can not pick it up right now.

I understand if you want to close it.  I'll make a separate note to try to revive it this spring. 

Hans

hjmjohnson:
@hansenms FYI:  I took a quick look, and most of this had already been included in other patches.

This updated pull request has much less content.  I'm only putting it here for historical purposes.

Feel free to close this pull request without merging it.

hansenms:
Let's close it now and open it back up when somebody is ready to lead it.

PR #57, opened 2016-07-09
Compression flags

hansenms:


hansenms:
@xueh2 Thoughts?

xueh2:
looks great to me.

PR #59, opened 2016-07-14
Icp prototype

vadre:


hansenms:
As I said, we will not be checking in 5000+ lines of code

PR #60, opened 2016-07-15
Icp2

vadre:
Updated EntityHeader
- replaced version with signature
- moved stream to the top position

Updated Entity class
- added virtual getters for stream, signature, entity type and storage type
- added getters to acquisition and image classes

Updated test_acquisitions


hansenms:
Looks good. We should also update the test for the image and I think there should be a function getVersion() on the entities that separate the signature part from the version part and just return the version. That can be added later. I will merge this now.

PR #61, opened 2016-07-19
Added virtual getVersion to Entity class. Made IsmrmrdHeader subclass…

vadre:
Added virtual getVersion to Entity class.
Made IsmrmrdHeader subclass of Entity, added virtual getters and constructor. 
Drafted Waveform class and methods. 
Added test_waveform and updated test_image and test_xml. 
Added bounds checking for image and NDArray classes per TODOs.

### DIFF
--- a/include/ismrmrd/xml.h
+++ b/include/ismrmrd/xml.h
@@ -8,6 +8,7 @@
 #define ISMRMRDXML_H
 
 #include "ismrmrd/export.h"
+#include "ismrmrd/ismrmrd.h"
 
 #include <cstddef>
 #include <new> //For std::badalloc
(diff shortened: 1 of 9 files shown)


include/ismrmrd/xml.h
```diff
@@ -319,7 +320,19 @@ namespace ISMRMRD
   };
 
   struct IsmrmrdHeader
+  : public Entity
```
hansenms: This is wrong. This ismrmrd header is the deserialized version of the XML header. Not the actual acquisitions. When the XML header is an "entity" on the wire, we need a different class for it or something. We should discuss.

tests/test_xml.cpp
```diff
@@ -10,6 +10,12 @@ using namespace ISMRMRD;
 const std::string XML_HEADER("\
 <?xml version=\"1.0\"?>\n\
 <ismrmrdHeader xmlns=\"http://www.ismrm.org/ISMRMRD\" xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xmlns:xs=\"http://www.w3.org/2001/XMLSchema\" xsi:schemaLocation=\"http://www.ismrm.org/ISMRMRD ismrmrd.xsd\">\n\
+	<EntityHeader>\n\
```
hansenms: This does not belong here. This relates to the transmission protocol and not the information content.

include/ismrmrd/xml.h
vadre: When should I stop by?

Vlad Dreyvitser [C] | Computer Programmer, NIH/NHLBI
Email: vlad.dreyvitser@nih.gov<mailto:vlad.dreyvitser@nih.gov>
Mobile: +1-703-835-3559
9000 Rockville Pike, Bethesda, MD 20814

PR #62, opened 2016-07-22
Icp2

vadre:
Removed entity header from Acquisition, Image, Waveform entities
Removed entity header and entity inheritance from IsmrmrdHeader
Added IsmrmrdText entity class and implementation.
Updated test routines for changed entities, and added new for the IsmrmrdText
Dataset class formatting changes only.

### DIFF
--- a/examples/protocol/main.cpp
+++ b/examples/protocol/main.cpp
@@ -12,226 +12,243 @@
 using boost::asio::ip::tcp;
 namespace po = boost::program_options;
 
-class Server {
-public:
-    Server(boost::asio::io_service& io_service, unsigned short port)//const tcp::endpoint& endpoint)
-      : io_service_(io_service),
-        port_(port)
-    { }
+class Server
+{
+  public:
 
-    void start()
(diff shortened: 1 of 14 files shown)


hansenms:
We need to walk through this together. I have several questions.

A couple of quick comments:
* It is not clear if Acquisition should be a subclass of Entity. Maybe it should, maybe it should not, but this needs to be discussed. I thought we had decided that it was not a subclass of entity.
* This PR contains edits to the protocol example code, that should maybe be in a separate PR. 
* There are *many* formatting changes that makes it hard to review the PR. There are so many line changes flagged but most of them are just formatting. I would prefer we do those changes in style/formatting PRs that don't change any actual code. When they are mixed like this review is hard.

vadre:
Please see below in orange

Thanks.

Vlad

Vlad Dreyvitser [C] | Computer Programmer, NIH/NHLBI
Email: vlad.dreyvitser@nih.gov<mailto:vlad.dreyvitser@nih.gov>
Mobile: +1-703-835-3559
9000 Rockville Pike, Bethesda, MD 20814

PR #70, opened 2016-09-02
Fix HDF5 detection on CMake 3.6.

ghisvail:
The [HDF5 find module](https://cmake.org/cmake/help/v3.6/module/FindHDF5.html) provided by CMake 3.6 breaks support for deprecated `HDF5_*` variables. This PR replaces the now unsupported `HDF5_C_INCLUDE_DIR` with the recommended `HDF5_INCLUDE_DIRS`.


ghisvail:
Link to bug report [downstream](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=835679).

sebastic:
It's better to support both the deprecated and recommended variables, because users may have version of CMake which doesn't support one or the other. I submitted the following patch to the above mentioned bugreport:
```patch
Description: Fix HDF5 include directory with CMake >= 3.6.0.
 HDF5_<LANG>_INCLUDE_DIR was deprecated some time ago,
 only HDF5_<LANG>_INCLUDE_DIRS is set for CMake >= 3.6.0.
Author: Bas Couwenberg <sebastic@debian.org>
Bug-Debian: https://bugs.debian.org/835679

--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,6 +95,9 @@ set(ISMRMRD_SOVERSION ${ISMRMRD_VERSION_
 find_package(HDF5 1.8 COMPONENTS C)
 
 if (HDF5_FOUND)
+    if(HDF5_C_INCLUDE_DIRS AND NOT HDF5_C_INCLUDE_DIR)
+        set(HDF5_C_INCLUDE_DIR ${HDF5_C_INCLUDE_DIRS})
+    endif()
     set (ISMRMRD_DATASET_SUPPORT true)
     set (ISMRMRD_DATASET_SOURCES libsrc/dataset.c libsrc/dataset.cpp)
     set (ISMRMRD_DATASET_INCLUDE_DIR ${HDF5_C_INCLUDE_DIR})
```

ghisvail:
@sebastic 

No offense, but the new variables were introduced back in 2009 (https://github.com/Kitware/CMake/commit/3b8cab3a4652a34819ee54c9753732f7d67c2102), that's long enough to drop usage of the old ones.

Plus we should not encourage our upstreams to use deprecated / non-supported variables. Perhaps the minimum CMake version should be bumped to whatever 2.8.x was back in 2009 indeed.

sebastic:
Since it's trivial to support both, that has my preference.

I'll leave it to the ismrmrd developers to decide which approach they prefer.

hansenms:
I don't see a reason to not support both.

ghisvail:
Also the [official documentation](https://cmake.org/cmake/help/v3.6/module/FindHDF5.html) says

```
This module will define the following variables:

HDF5_FOUND - true if HDF5 was found on the system
HDF5_VERSION - HDF5 version in format Major.Minor.Release
HDF5_INCLUDE_DIRS - Location of the hdf5 includes
HDF5_INCLUDE_DIR - Location of the hdf5 includes (deprecated)
HDF5_DEFINITIONS - Required compiler definitions for HDF5
HDF5_LIBRARIES - Required libraries for all requested bindings
HDF5_HL_LIBRARIES - Required libraries for the HDF5 high level API for all
                    bindings, if the HL component is enabled
```

and
```
For each enabled language binding, a corresponding HDF5_${LANG}_LIBRARIES variable will be defined.
```

nothing about `HDF5_${LANG}_INCLUDE_DIRS`. Sure `HDF5_C_INCLUDE_DIRS` does work **now**, but who knows whether it will stay around in the future or disappear in a future CMake update as it did for 3.6 with `HDF5_C_INCLUDE_DIR`.

Your call.

hansenms:
@ghisvail, we can close this, right? It should not bee needed anymore or am I wrong?

ghisvail:
I believe upstream CMake actually fixed the regression. Closing.

PR #74, opened 2016-11-04
support building against HDF5 1.10

### LINKED ISSUE #73: FTBFS with HDF5 1.10 due to a change of fileid type
Just forwarding an [issue](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=841959) we had with compilation of `ismrmrd` on Debian following the transition of the archive to hdf5 v1.10.

Essentially, it all comes down to a change in the HDF5 API regarding the returned type of `H5Fopen`, which is no longer an `int` but a `long long` in v1.10. Actually, HDF5 encapsulates this return type to a `hid_t`, and the patch proposed by Gilles uses this type for `ISMRMRD_Dataset.fileid` which eventually fixes the problem.

Are you happy with this change? If so, I can forward Gilles' patch in a subsequent PR.

ghisvail:
Closes #73

### DIFF
--- a/include/ismrmrd/dataset.h
+++ b/include/ismrmrd/dataset.h
@@ -9,6 +9,7 @@
 #define ISMRMRD_DATASET_H
 
 #include "ismrmrd/ismrmrd.h"
+#include <hdf5.h>
 
 #ifdef __cplusplus
 #include <string>
@@ -28,7 +29,7 @@ extern "C" {
 typedef struct ISMRMRD_Dataset {
     char *filename;
     char *groupname;
-    int fileid;
+    hid_t fileid;
 } ISMRMRD_Dataset;
 
 /**


hansenms:
@ghisvail this BTW as usual caused major downstream headaches. Because <hdf5.h> is now included in the ismrmrd.h file, all applications need the HDF5 include dir when compiling. Major headache.

ghisvail:
> all applications need the HDF5 include dir when compiling.

Indeed. This can be solved by adding the HDF5 detection to `FindIsmrmrd.cmake` and appending HDF5_INCLUDE_DIRS and HDF5_LIBRARIES to their ISMRMRD counterparts. I will have a look.

ghisvail:
Meanwhile, feel free to revert.

ghisvail:
The use of export targets in the modernization proposed in #56 would greatly help actually. What is currently blocking this PR?

hansenms:
@ghisvail I have put in fixes in the downstream projects, but an upstream fix in ISMRMRD would be nice. We don't use FindIsmrmrd.cmake anymore

@hjmjohnson said to hold off on #56. I will let him comment.

ghisvail:
> We don't use FindIsmrmrd.cmake anymore

What do you use now then? A [config-file](https://cmake.org/cmake/help/v3.0/manual/cmake-packages.7.html#id7) CMake package? The latter would be ideal.

hansenms:
Yes, config file.

ghisvail:
> Yes, config file.

Where? I can't see any `IsmrmrdConfig.cmake` or `ismrmrd-config.cmake` being generated by `ismrmrd` 1.x or 2.x, only `FindIsmrmrd.cmake` is provided so far. Or am I missing something?

I also looked at the fix in [Gadgetron](https://github.com/gadgetron/gadgetron/commit/354f257fce9c9e52ca9723a4cbbd8ea73c35e446). That's fine as a temporary fix, but I agree with you that this should be fixed upstream.

In the ideal case, we'd want to allow `Ismrmrd` to be detected and used as an imported target (which #56 should enable) and have HDF5 declared in the public interface of `Ismrmrd` via `target_include_directories` and `target_link_libraries`. This way doing `target_link_libraries(Gadgetron, Ismrmrd)` will propagate the dependency on HDF5 without requiring detection.

Without the imported target mechanism, the other options is to not use `hid_t` (and therefore avoid a public dependency on HDF5) and just upgrade to ISMRMRD_Dataset.fieldid to `long long`. I can test that too.

hansenms:
https://github.com/ismrmrd/ismrmrd/blob/master/cmake/ISMRMRDConfig.cmake.in

ghisvail:
Thanks @hansenms, that's what I was missing. Since Debian only tracks tagged releases, I assumed `1.3.2` was the latest stable version of `ismrmrd`. However based on your link, it appears that `1.3.3` exists (with config-mode detection). Could you tag it please, so I can update the packaging?

With what's in `1.3.3` it should be possible to just add `HDF5_INCLUDE_DIRS` and `HDF5_LIBRARIES` to their ISMRMRD counterparts (assuming #70 is merged in) so that the HDF5 dependency gets propagated downstream via `IsmrmrdConfig.cmake`.

ghisvail:
Looks like I got it figured out. I ran a build instance for the packaging of `1.3.3` and successfully [tested](http://debomatic-amd64.debian.net/distribution#unstable/ismrmrd/1.3.3-1/autopkgtest) the example against the packaged binaries (which covers detection via `find_package`, linkage and execution).

I'll tidy up the patches and roll up some pull-requests. Thanks for your help on this.

hansenms:
Thanks, let me know when you have something.

PR #78, opened 2017-01-05
Update serialize.m

TWhelan3:
Change append_node to be able to append vectors of values.


TWhelan3:
The members of SequenceParameters are mostly vectors, but serialize.m does not treat them like that.

PR #79, opened 2017-01-11
[CMake] insert STATUS in front of messages

KrisThielemans:
This means the message will end-up in the log as expected.


KrisThielemans:
this prevents a problem with ccmake blocking the screen etc.

Note that gadgetron has its own FindFFTW3.cmake. This would have to be fixed as well. Better is probably to remove that, and use ismrmrd's (after install it is now available in the same path as FindISMRMRD.cmake).

hansenms:
Gadgetron has to have its own FindFFTW3.cmake since Gadgetron can be compiled without ISMRMRD.

PR #81, opened 2017-02-04
[CMake] fixed handling of spaces in folders

KrisThielemans:
On Windows, the install target-directory often contains spaces. so we need quotes around the variables such that an import doesn't fail. We've tested this on Mac/Linux and Windows.


hansenms:
I'm inclined to say that people should not install in folders with spaces in it because it is asinine but if we can help without breaking anything else, I suppose it is OK. I will merge.

PR #89, opened 2017-04-24
Use HDF5_INCLUDE_DIRS as opposed to HDF5_C_INCLUDE_DIR

KrisThielemans:
necessary for new CMake (tested with CMake 3.7.2 on Lubuntu 15.10)


hansenms:
Can't really evaluate the cross-platform impact of this right now. May well break stuff on Mac or Windows. Lubuntu 15.10 is not a reference platform and this "fix" is not needed on Ubuntu 16.04 (which is a reference platform).

KrisThielemans:
Agree that this is difficult to test. I should have mentioned however that this fix should be backwards compatible.

When I contributed a fix to FindHDF5.cmake (necessary for Windows) to CMake, Brad King (Kitware) said

> The HDF5_${lang}_INCLUDE_DIR variables are not needed for compatibility because we never documented that they were provided by earlier versions.  Only HDF5_INCLUDE_DIRS is documented.

See https://gitlab.kitware.com/cmake/cmake/merge_requests/463

This implies that we should only use HDF5_INCLUDE_DIRS. 

With this fix, ismrmrd builds on CentOS 7 (CMake 3.7.0), Lubuntu 15.10 (standard CMake and 3.7.2), Ubuntu 16.04 (standard CMake and 3.7.2), and will work with CMake 3.9 on Windows. It also works on Ubuntu when you compile your own HDF5 (current version doesn't). I can get you MacOS Sierra tomorrow (but isn't there a test-build already?).

PR #90, opened 2017-04-26
Modification to include/dataset.h

DrinkJava:
Added a template definition for readNDArray(), which without was causing a segmentation violation when called from data processing wrapper.


hansenms:
Not quite sure that this is. What problem does this solve and why are there two commits?

PR #98, opened 2017-10-17
Vectorise slow Matlab "for" loop

ojosephs:


xueh2:
looks good to me!
@ojosephs, when will the big Matlab PR come in? I am waiting to organize a show for you .

PR #103, opened 2018-03-23
User string field in trajectory description and fractional acceleration factors

DietrichBE:
- Added user string to trajectory description in XML header for additional non numeric descriptive data, such as for example a trajectory basis type. For example "spherical harmonic basis" or any other descriptive information. It appears as if the user string type was simply forgotten in the trajectory description field?

- Changed the type of the accellerationFactor fields in the XML schema to double, in order to comply with for example Philips SENSE scan, which allow and commonly use fractional acceleration factors such as 2.3.

### DIFF
--- a/schema/ismrmrd.xsd
+++ b/schema/ismrmrd.xsd
@@ -182,6 +182,7 @@
       <xs:element maxOccurs="1" minOccurs="1" name="identifier" type="xs:string"/>
       <xs:element maxOccurs="unbounded" minOccurs="0" name="userParameterLong" type="userParameterLongType"/>
       <xs:element maxOccurs="unbounded" minOccurs="0" name="userParameterDouble" type="userParameterDoubleType"/>
+      <xs:element maxOccurs="unbounded" minOccurs="0" name="userParameterString" type="userParameterStringType"/>
       <xs:element maxOccurs="1" minOccurs="0" name="comment" type="xs:string"/>
     </xs:sequence>
   </xs:complexType>
@@ -243,8 +244,8 @@
 
   <xs:complexType name="accelerationFactorType">
     <xs:all>
-      <xs:element name="kspace_encoding_step_1" type="xs:unsignedShort"/>
-      <xs:element name="kspace_encoding_step_2" type="xs:unsignedShort"/>
+      <xs:element name="kspace_encoding_step_1" type="xs:double"/>
+      <xs:element name="kspace_encoding_step_2" type="xs:double"/>
     </xs:all>
   </xs:complexType>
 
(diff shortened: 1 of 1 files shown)


dchansen:
Hi Dietrich,

I am sorry that this has been ignored for so long. For whatever reason, I wasn't getting notifications from ISMRMRD.
Changing the accellerationFactor to double seems like a reasonable change, although I will need to check what impact it would have on downstream projects.
User strings were intentionally left out  of the trajectory description. We would much rather have  additional fields with specific names. 
Otherwise, ISMRMRD files will not really be compatible between different facilities, as everyone would use different meta fields.

DietrichBE:
Hi David,
Sorry for the late response. I also missed somehow the message from github.
- Changing the acceleration factor to double would be great.
- Regarding the trajectory description I agree, it makes more sense to have some specific fields. We would need the following fields: 
- - kBasisType: this should be a sort of general identifier for what kind of spatial basis the trajectory uses. Examples would be "sphericalHarmonic" or "sphericalHarmonicsAndSecondOrderSymmetricGradientConcomitant". Thats kind of what we use typically.
- - kBasisLabels: An basis function identifier for each trajectory dimension, such as for example "B0, X, Y, Z, XY, ZY, 3Z^2-R^2, ..."
- - kBasisMask: similar to the channel mask, indicating selected basis functions.
- - kBasisUnits: Describing the units of each trajectory dimension. Example: "rad, rad/m, rad/m, rad/m, rad/m^2, ..."
- - And maybe an additional string field describing which method was used to calculate/generate/measure the trajectory

What do you think? What would be the next step? I could make an example xml schema.

Cheers,
Benjamin

hansenms:
Just picking this up. This PR has been sitting around for some time, so we should either close it or get it dusted off.  Some comments:

1. Acceleration factor for parallel imaging should *not* be a non-integer. What Philips actually does is that they encode a larger FOV, which is then cropped at the end. So for say acceleration factor 2.3, it is really a bigger FOV and sampling every third line and the recon solves for 3 aliases. So I would recommend being explicit with that. You could achieve the same fractional SENSE factor by having phase oversampling and acceleration factor 3. I would suggest keeping it as an integer. 
2. I actually think text descriptors for the trajectory are "fine". But I would recommend instead just using different identifiers. So you could call your trajectory `spiral:come-sub-type` and then have parameters for that type. We should not attempt to add named parameters for every trajectory type. There are just too many.

So I think overall, this PR is probably not needed, but I could be convinced otherwise. If we do have it, we should make sure the hash for the schema is updated (build will fail without it) and the minor version has to be bumped up. 

@dchansen, thoughts?

hansenms:
I'm gonna close this one, @dchansen @DietrichBE please reopen or start new issue as appropriate if my comments above do not cover it.

PR #109, opened 2018-08-18
CMake: re-enable advanced version checking features

Takishima:
The changes proposed here are compatible with all major versions of CMake since CMake 2.8.12 which is now 5 years old (Ubuntu 14.04 or Debian jessie required).


dchansen:
Hi Takishima,

I  am sorry this has been sitting in the pull request queue for so long. That is obviously an oversight on my part. 
I think we can safely move to CMake 3.0+ now.

PR #123, opened 2019-01-25
Fix some MetaContainer inheritance and const issues

mklassen:
- make typedef map_t protected instead of private because protected variable of type map_t exists.
- make empty() const as it does not change the object.


dchansen:
Thanks for taking the time to commit these things. ISMRMRD was not really written with const correctness in mind, but we're in the process of fixing that.

PR #140, opened 2019-12-07
Changed number of user_float from 8 to 16

chcodes:
Dear ISMRMRD,

The Siemens VD software version allows for 24 arrays of custom user defined data of datatype uint16_t (see line 141 in /siemens_to_ismrmrd/siemensraw.h). During the conversion of the raw Siemens data file to the ISMRMRD format, these data arrays get stored either as int32_t or float (see lines 1029 to 1046 in /siemens_to_ismrmrd/main.cpp). However, only the first 16 arrays are stored (the first 8 as int32_t and the second 8 as float); the last 8 arrays of data are not stored.

If what I have described is correct, I propose to store these last 8 arrays of data as float's in the ISMRMRD file. I have hopefully made all the required changes to both the ismrmrd and siemens_to_ismrmrd repositories, and request that you pull these changes. Our group stores peripheral monitoring unit data in these last 8 arrays, and have tested that these changes do successfully write this data to the ISMRMRD file, which was then read in by both Matlab and Python. I can provide Siemens raw data files with this data if you require.

Regards,
Chris Huynh


hansenms:
This renders this a new (binary incompatible) version and should **not** be merged.

dchansen:
Yes, this will break backwards compatibility. We can consider it for a new major version of ISMRMRD.

PR #154, opened 2020-05-04
Add check for critical environment variables.

roopchansinghv:
Put in check to stop cmake configuration and print (hopefully useful) error message.


roopchansinghv:
Thanks @dchansen for incorporating this!

Will probably do one more pull request to sync up changes in the master branch, so they are fully integrated into the GE-compat.

PR #167, opened 2020-12-22
Fix PugiXML variable name

LocutusOfBorg:
This was making the testsuite fail because with new PugiXML the find_package can't be done twice anymore
See: https://github.com/zeux/pugixml/issues/393


hansenms:
Fixed in #169. Closing this one.

PR #168, opened 2021-02-14
Create ismrmrd_build.yml

hansenms:


hansenms:
@dchansen, this bootstraps a simple build pipeline on the repo. It is not complete yet, but it builds and runs the tests. We can iterate from there to get docs built and pushed.

PR #169, opened 2021-02-16
Removed sphinx docs build, treating warnings as errors

hansenms:
This PR removes the sphinx docs generation and treats the main README.md as the entry point for documentation. Other changes include cleaning up some compiler warnings and treating warnings as errors going forward.


dchansen (approved):
Looks good. We should consider having ndarray initialize with "new", and be a nice standards conforming container.

PR #134, opened 2019-06-22
fix boost::program_options linking error

KrisThielemans:
On Windows, I need `BOOST_ALL_NO_LIB` to prevent double linking errors and `BOOST_ALL_DYN_LINK` to prevent linking errors due to missing variables ` boost::program_options::arg` and `boost::program_options::options_description::m_default_line_length`.

Tested with Boost 1.69.0, VS2015 and CMake 3.13.2.

PR #170, opened 2021-02-16
Added windows workflow

hansenms:


dchansen:
Looks good to me. Is anything missing?

hansenms:
@dchansen almost there, I think. Just making sure that the pipeline actually fails when it should. So, a few more iterations and I will publish the PR in not too long and let you know. Good thing is that we do have a Windows pipeline.

hansenms:
OK @dchansen this should be ready. And @KrisThielemans this includes #134 so we can close that when this is merged.

cmake/FindFFTW3.cmake
```diff
@@ -75,17 +75,16 @@ set(_check_list)
...
-      find_library(${_LIB}_LIBRARY NAMES lib${_lib}-3 lib${_lib}
+      find_library(${_LIB}_LIBRARY lib${_lib}-3 ${_lib}
         HINTS $ENV{FFTW3_ROOT_DIR} PATH_SUFFIXES lib)
       mark_as_advanced(${_LIB}_LIBRARY)
       list(APPEND FFTW3_LIBRARIES ${${_LIB}_LIBRARY})
       list(APPEND _check_list ${_LIB}_LIBRARY)
     endforeach()
 
-    message(STATUS "FFTW3 WINDOWS libraries: " ${FFTW3_LIBRARIES})
+    message("FFTW3 WINDOWS libraries: " ${FFTW3_LIBRARIES})
```
KrisThielemans, line 87: just checking, why do you remove the `STATUS`? From the CMake doc:
```
(none) or NOTICE
Important message printed to stderr to attract user’s attention.

STATUS
The main interesting messages that project users might be interested in. Ideally these should be concise, no more than a single line, but still informative.
```

hansenms: Yeah, no reason really. I was coping a section from the Gadgetron repo that does this a bit differently. I put it back in.

hansenms:
@dchansen are OK to get this merged so we have some coverage on the Windows side before proceeding?

dchansen (approved):

PR #172, opened 2021-03-17
Updated Optional to be (more) compatible with std::optional

dchansen:


hansenms:
Can this be closed now?

PR #173, opened 2021-04-12
Updated schema type names

dchansen:


hansenms:
So since this changes the schema the minor version has to be bumped and the hash updated in the cmake file.

hansenms (approved):

PR #174, opened 2021-04-20
First version of flexible header documentation

wuerflts:


hansenms:
@wuerflts apologies, looks like I didn't type up my comments. The table look nice, no worries there. My main concern is how do we become aware of when the table should maybe be updated? My suggestion would be that somewhere in the document with the table we hide the hash of schema and if the scheme changes, at a minimum the hash should be updated. We actually do this in already in the CMakeLists.txt to ensure that the minor version is updated. Something that makes you stop and think did I really update the table if I updated the schema.

hansenms:
@wuerflts closing per our discussions. Please reopen against the new readthedocs infrastructure once it is up and running.

PR #175, opened 2021-05-07
Added stationID

dchansen:
This pull request adds the concept of a "stationID" to ISMRMRD. 
The purpose is to be able to identify the scanner of a particular dataset. This is important for scanner specific calibrations, such as GIRF, where we want to later retrieve a scanner specific calibration.


include/ismrmrd/xml.h
```diff
@@ -219,6 +219,7 @@ namespace ISMRMRD
     std::vector<CoilLabel> coilLabel;
     Optional<std::string> institutionName;
     Optional<std::string> stationName;
+    Optional<std::string> stationID;
```
hansenms, line 222: nit: deviceID

hansenms (approved):

PR #176, opened 2021-05-14
Update README.md: MacPorts installation

eborisch:
Document that MacPorts can be used to install ismrmrd directly on macOS. Pre-compiled binaries are used to perform the installation so long as the default MacPorts installation options (location & permissions) are used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,16 @@ sudo make install
 This will install the library in `/usr/local/` by default. To specify
 an alternative installation directory, pass `-D CMAKE_INSTALL_PREFIX=<install dir>` to `cmake`.
 
+### macOS installation
+
+ISMRMRD and its dependencies can be installed via MacPorts (https://macports.org) with:
+
+```
+sudo port install ismrmrd
+```
+
+This will install the library, binaries, documentation, and examples into `/opt/local` by default. Changing the default MacPorts directory is not recommended as it prevents using pre-compiled packages.
+
 # Format Details
 
 The raw data format combines a mix of flexible data structures (XML header) and fixed structures (equivalent to C-structs). A raw data set consist mainly of 2 sections:


hansenms:
I don't think Macports is being maintained, so I don't think we should advertise that as a way to install.

eborisch:
Perhaps I should have been clearer; I maintain the ismrmrd (currently at the latest 1.5.0) port on MacPorts, which as a project is also very active and maintained. Pre-compiled binaries are used for the install (https://packages.macports.org/ismrmrd/) so it's also quite fast to install. (Activating boost takes most of the time.)

hansenms:
@eborisch thanks for that context. Yes it does make a difference and would have been great context in the PR. That said, there are a number of ways to obtain ISMRMRD (compile yourself, vcpkg, debian packages, rpms, conda packages, and now macports) and some of the instructions (including compile instructions the README) have not been maintained well. We are in the process of cleaning up the documentation and moving it to readthedocs.org. When we do that I suggest a reworked page on stall instructions and we can add this to that. What I would like in addition to the instructions is who/how to ping when it is out of date, or there are problems. There will undoubtedly be an issue at some point and if the packages are not being maintained on by the group maintaining this repo, we should have some webpage/contact info.

eborisch:
The output of `port info ismrmrd` contains contact information. All MacPorts ports are managed in the same (active, over 100 commits in the past week) repository (https://github.com/macports/macports-ports/) where pull requests can be made, as well as a TRAC instance with tickets that can be generated if a PR isn't available: https://trac.macports.org/wiki/Tickets.

While I work in MR, I haven't personally used ISMRMRD (beyond at a hackathon); I was primarily trying to make it easier for others to install if they wanted to kick the tires. If you think an additional installation method just adds to the documentation and support overhead from the project's perspective, feel free not to include it.

hansenms:
@eborisch i certainly appreciate the effort. I’m just trying to figure out what happens when it doesn’t work (it will break at some point). If it is “advertised” on the README without some disclaimer that it is not maintained by the people maintaining this repo, then I think people will try to get help for it here. I would. So let’s take it into to the right place of the documentation with the right contact information.

Hope that makes sense.

PR #179, opened 2021-10-07
Fix typos found by codespell

DimitriPapadopoulos:


hansenms:
@DimitriPapadopoulos can we update this by merging latest master branch. You would also need to increment the micro version in the CMakeLists.txt file.

DimitriPapadopoulos:
Increment this version?
https://github.com/ismrmrd/ismrmrd/blob/c378c5b099f0f56e7de3f4222a75aaba63a6ad1a/CMakeLists.txt#L1

hansenms:
This version:
https://github.com/ismrmrd/ismrmrd/blob/master/CMakeLists.txt#L84

The patch version of the library. We could say it isn't necessary here, but the UI (the help text) of one of the applications changes, so maybe bump the patch version.

CMakeLists.txt
```diff
@@ -80,7 +80,7 @@ endif ()
 #in the utility libraries, that don't affect the data format itself.
 # For more information see http://semver.org/
 set(ISMRMRD_VERSION_MAJOR 1)
-set(ISMRMRD_VERSION_MINOR 5)
+set(ISMRMRD_VERSION_MINOR 6)
```
hansenms, line 83: This should be the PATCH version you update. The guidance is anything that messes with the binary layout of structures is a major version. Anything that messes with the flexible (XML) header is a minor version. And small tweaks, like this one, is PATCH. So it would be 1.5.1.

hansenms (changes requested):

hansenms (approved):

PR #185, opened 2022-02-02
Added support for automatic RPM builds

dchansen:
This pull request fixes the default install directory and adds a spec file which allows for automatic RPM builds on [copr](copr.fedorainfracloud.org/) (like [this](https://copr.fedorainfracloud.org/coprs/davidchansen/ISMRMRD/)).

### DIFF
--- a/.github/workflows/rpkg_build.yml
+++ b/.github/workflows/rpkg_build.yml
@@ -1,0 +1,28 @@
+name: RPKG
+
+on:
+  push:
+    branches:
+      - master
+  
+  pull_request:
+    branches:
+      - master
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    
+    container:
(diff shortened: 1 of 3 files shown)


hansenms:
Should we exercise this in the build pipeline?

dchansen:
I've added a github action that runs the spec file and makes sure that it actually builds.
For hosting the repo, COPR has [support](https://docs.pagure.org/copr.copr/user_documentation.html#github-webhooks) for Githubs webooks.
Unfortunately, a project is tied to a single user, so we should either make an "offical" ISMRMRD Fedora account, or use one of our own. Long term, I think getting ISMRMRD into Fedora is the way to go.

hansenms (approved):
Looks. At some point, we may want to have these package builds in another repo, but for now this is a good enough place.

PR #189, opened 2022-02-07
PPA support

dchansen:
This pull request adds support for auto-building pacakges for debian-based distributions (such as Ubuntu) on launchpad.
In particular it enables us to have a custom ppa like [this](https://launchpad.net/~gradient-software/+archive/ubuntu/experimental)
for ISMRMRD.


.github/workflows/rpkg_build.yml
```diff
@@ -0,0 +1,28 @@
+name: RPKG
```
hansenms, line 1: Not totally sure why this file is listed as a change here . You may need to merge master or something? This file is in master already.

debian/copyright
```diff
@@ -0,0 +1,43 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: ismrmrd
+Upstream-Contact: <preferred name and address to reach the upstream project>
+Source: <url://example.com>
```
hansenms, line 4: Some boiler plate in this file that would need to be replaced, e.g. license, etc.

debian/rules
```diff
@@ -0,0 +1,22 @@
...
+# package maintainers to append CFLAGS
+#export DEB_CFLAGS_MAINT_APPEND  = -Wall -pedantic
+# package maintainers to append LDFLAGS
+#export DEB_LDFLAGS_MAINT_APPEND = -Wl,--as-needed
+
+
+%:
+	dh $@
+
+# For some reason, I'm missing the dh_strip_nondeterminism script on Fedora.
```
hansenms, line 20: Why is that relevant?

hansenms (changes requested):
Made some comments inline. Two main comments would be:

1. Do we need to exercise this in the build pipeline? I think probably yes.
2. Some boiler plate stuff to tidy up.
3. Should we be pushing these packages somewhere?

hansenms:
This has been sitting for a bit, so I am closing. Please re-open if it is still relevant.

PR #192, opened 2022-02-12
Fix more typos, including a function name (sensititivies → sensitivities)

DimitriPapadopoulos:
This obviously breaks compatibility. How do you want to handle this? Keep the old function and have it call the new function? This will requite a minor version bump.


hansenms:
Yes, this is an API breaking change of sorts. It is not super critical since it is not in the core library, but in some of the tools. I would suggest bumping the minor version on this and we can take this change.

hansenms (approved):

PR #194, opened 2022-02-15
Add enums for RGB image types and phase stabilization flags

kspaceKelvin:
Added the following:

- ``ISMRMRD_IMTYPE_RGB`` for color images, where images have 3 channels representing red, green, and blue channels.
- ``ISMRMRD_ACQ_IS_PHASE_STABILIZATION`` and ``ISMRMRD_ACQ_IS_PHASE_STABILIZATION_REFERENCE`` for readouts associated with gradient echo phase stabilization.


hansenms:
I think this is at a minimum a patch version change, but probably a minor version change, since it expands the API.

The topic of RGB images has come up a few times, but I think I forgot what we settled on. Is this the entirety of what is needed for RGB images? @xueh2 where did we land on this?

kspaceKelvin:
Ah, you're right.  I'll update the minor version.

This is the entirety of what is required for RGB images in terms of enum changes.  In practice, there is more to the "contract" here.  Notably, the convention that I've used is that RGB images use data_type ISMRMRD_USHORT with 3 channels, representing the red, green, and blue channels of the RGB image. Image data values are in the range 0-255 (8-bit color depth).  These are up for discussion though and would be described in the documentation to follow.

@hansenms Thoughts on the failed HDF5 dependency in the vcpkg Windows build?

hansenms:
I am a bit concerned about taking stuff in that is part of some functionality, but we don't have a clear write-up on what the totality of the implementation will be. It may or may not be that this will be the only flags we will need for RGB support, but I don't think we should start rolling in PRs with some of the stuff unless we have spec'ed out what the final solution would look like. I know this is a departure from how things have been done, but please bear with me, it is worth it. Can we jot down (in a hackmd.io doc or whatever) what RBG support will looks like? Why are we doing it? Who is it for? What will the implementation look like? How will we test it?

As for the failing build, it may just be a fluke (failure to pull a package), it just ran on another PR, so I don't think it is a problem.

hansenms:
I would suggest splitting the RBG stuff out from the acquisition flags (which are just “informational labels”), unless we have a write up (not a book, just a page) of how we will do RGB.

kspaceKelvin:
I'm up for more documentation/testing, but I'm not sure how much more there is to describe beyond the above.  The "why" is so that we can support color images (e.g. figures, renderings, reports) without the need for workaround colormaps.  I'm not sure what kind of implementation or testing would belong here -- would you simply want a test case added to ``test_images.cpp`` which sets 3 channels and ISMRMRD_USHORT?  I see that more of an example of how we want it to be done, as this doesn't validate that anyone else's data follows this guidance.

hansenms:
What makes this different from the acquisition flags is that the underlying implementation is not settled. So specifically, if we say it is an image with 3 channels, then there should be some validation or something. That you cannot have that image type without having 3 channels? Should it be 3 or 4 (if we support mask?). Or it is actually that the individual pixels are tuples of 3 (or 4) values and there is 1 channel in the image. So my hesitation here is that this flag comes with a lot of semantic baggage and without settling that we are setting ourselves up for a bunch of hacks (which are already happening). 

So what I would like to see for this flag is a clear definition of how it is supposed to be interpreted and then some kind of test that validates that you cannot have that set without having 3 channels if that is how we define it. Adding the flag without that is postponing the decision on what the flag means. 

Hope that makes sense.

kspaceKelvin:
Let's defer the RGB discussion to another PR then.  This PR has been revised to include only the phase stabilization flags.

hansenms (approved):

PR #196, opened 2022-02-20
CMake modernisation and fixes

### LINKED ISSUE #112: missing hdf5.h due to wrong CMakeLists.txt
f51e8469c3d219dbe85d42a0dcd791f2ac751368 changed 
```
-   set (ISMRMRD_DATASET_INCLUDE_DIR ${HDF5_INCLUDE_DIRS})
-   set (ISMRMRD_DATASET_LIBRARIES ${HDF5_LIBRARIES})
+  set (ISMRMRD_DATASET_INCLUDE_DIR ${HDF5_INCLUDE_DIR})
+  set (ISMRMRD_DATASET_LIBRARIES ${HDF5_C_LIBRARIES})
```
and similarly substituted `HDF5_INCLUDE_DIRS` with `HDF5_INCLUDE_DIR` elsewhere This causes problems on some systems and is incorrect. It reverts an old bug, see e.g. [a debian bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=835679) and the corresponding PR #70. This PR was abandoned at the time, apparently due to changes in CMake, but the problematic behaviour is now there (Tested with CMake 3.31.1 on Ubuntu 16.04). The bugreport says it all. The CMake doc says
```
HDF5_INCLUDE_DIRS - Location of the hdf5 includes
HDF5_INCLUDE_DIR - Location of the hdf5 includes (deprecated)
```

### LINKED ISSUE #183: find_dependency(HDF5) in config could be conditional
I believe that these lines should be included only if `USE_HDF5_DATASET_SUPPORT=ON`
https://github.com/ismrmrd/ismrmrd/blob/feb3040f5d65148a45d84fd0125091a41138a2c6/cmake/ISMRMRDConfig.cmake.in#L44-L49

If CMake version >= 3.9, it doesn't matter too much as `find_dependency` [will forward](https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html) the `REQUIRED`. So the worst thing that can happen is that we will be searching for HDF5 when it actually isn't used.

If CMake version < 3.9, the `find_package` statement should have a `REQUIRED` when  `USE_HDF5_DATASET_SUPPORT=ON`.

### LINKED ISSUE #184: need to use target_include_directories for HDF5 (and others)
https://github.com/ismrmrd/ismrmrd/blob/4c453a2a5596023b21931b6c5ad01201ceb6acf3/CMakeLists.txt#L139-L140
should use `target_include_directories`, such that dependency on HDF5 gets sorted for anyone using ISMRMRD.

KrisThielemans:
Fixes #184 
Fixes #183 
Fixes #112 

Note that I bumped CMake requirement to 3.0. Should work with that I believe, but I only tested with 3.17


hansenms:
Thank @KrisThielemans The failing build here, I believe is due to a missing apt-get update. There is a fix for that in #195 so we will merge that first.

hansenms:
@KrisThielemans I have merged master on this an bumped the patch version. It all looks good. One question. Have you tried building (the latest) gadgetron against this version and does it "just work". Or are there changes needed on the Gadgetron CMake configurations?

KrisThielemans:
> Have you tried building (the latest) gadgetron against this version and does it "just work". Or are there changes needed on the Gadgetron CMake configurations?

Unfortunately I still cannot test gadgetron master without going through a lot of pain (for which I have no time). Hopefully someone else can try this, but I expect no change is necessary.

I did try a standalone install of the `examples/c` (Possibly that isn't tested at the moment?), which is how I discovered the problem with `ISMRMRD_LIBRARIES=ismrmrd` in the `ISMRMRDConfig.cmake` (. After that fix, it worked out of the box.

hansenms:
I just tried building the siemens converter against this branch and when running cmake, I get this warning:

```
CMake Warning (dev) at /anaconda/envs/gadgetron-ismrmrd/lib/cmake/ISMRMRD/ISMRMRDConfig.cmake:68 (if):
  if given arguments:

    "ON"

  An argument named "ON" appears in a conditional statement.  Policy CMP0012
  is not set: if() recognizes numbers and boolean constants.  Run "cmake
  --help-policy CMP0012" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.
Call Stack (most recent call first):
  CMakeLists.txt:70 (find_package)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

I *think* this comes from the conditional around dataset support that you added and should be fixed.

hansenms:
> 
> Unfortunately I still cannot test gadgetron master without going through a lot of pain (for which I have no time). Hopefully someone else can try this, but I expect no change is necessary.
> 

I would really recommend trying out the conda environments for this. It is a great way to have multiple different versions and configurations side by side and avoid some of the pain that none of us really have time for.

KrisThielemans:
> > Unfortunately I still cannot test gadgetron master without going through a lot of pain (for which I have no time). Hopefully someone else can try this, but I expect no change is necessary.
> 
> I would really recommend trying out the conda environments for this. It is a great way to have multiple different versions and configurations side by side and avoid some of the pain that none of us really have time for.

yeah I know. Sadly I struggle whenever I try some `conda` stuff. This one prevented me for about 1 year to do anything at all https://github.com/conda/conda/issues/8051. Now I'm struggling with
https://github.com/SyneRBI/SIRF-SuperBuild/issues/655#issuecomment-1047195429.

Also, once you go `conda`, you really have to go all the way (i.e. use conda-build etc) in my experience. As it wraps compilers etc, conflicts with shared libraries will happen very quickly.

Anyway, off topic.

KrisThielemans:
Regarding [CMP0012](https://cmake.org/cmake/help/latest/policy/CMP0012.html), this will happen when `cmake_minimum_required`  is older than 2.8. This is indeed the case for the converter
https://github.com/ismrmrd/siemens_to_ismrmrd/blob/b34196bd370c1aec4f930e9fbcccafd7c17d681b/CMakeLists.txt#L1
Are you sure you really want to support such old CMake versions? Are you testing it?

We can work around it here. I'd then suggest to do something like
```cmake
set (ISMRMRD_USE_HDF5_DATASET_SUPPORT @USE_HDF5_DATASET_SUPPORT@)
if (ISMRMRD_USE_HDF5_DATASET_SUPPORT)
 ...
endif()
```
what do you think?

hansenms:
@KrisThielemans I think you are right here, we should bump the min cmake version for `siemens_to_ismrmrd` to 3.0 instead of doing something convoluted here that we don't really test anyway.

I have built the Gadgetron against this branch and it all seems to work with all tests passing. 

Lemme take another look here and run it by the group and I will try to get this merged today (and close 3 issues!). Thanks of for efforts here.

hansenms (approved):

PR #197, opened 2022-02-25
Add missing close() statements to Dataset.m

thulrich:
Add missing close() statements to Matlab code. According to the HDF5 documentation, this should be done to free resources used by the respective objects.
See e.g. https://portal.hdfgroup.org/display/HDF5/H5P_CLOSE


kspaceKelvin:
I agree -- I think these were just accidentally omitted in the original code.  @hansenms Do you know if we can trigger the MATLAB test check for this PR even though the check was created afterwards?

hansenms:
@kspaceKelvin think that did it.

kspaceKelvin (approved):

Issue #181, opened 2022-01-12
MATLAB: Some valid ismrmrd headers can't be represented in Matlab


kristofferknudsen:
Some of the choices made in ismrmrd.xml.(de)serialize means some otherwise valid headers can't be represented in Matlab. 

The Matlab functions model the ismrmrd header as a nested tree of structs and arrays. This is normally fine, but causes issues where char arrays or structures with different attributes needs to share array. 

I've prepared a few unit tests, to illustrate the issues: 
```matlab
function testCanReadExtendedExample(test)
    header = ismrmrd.xml.deserialize(test.TestData.headers('ismrmrd_example_extended.xml'));

    written = ismrmrd.xml.serialize(header);
    written = ismrmrd.xml.deserialize(written);
    
    test.assertEqual(written, header, 'RelTol', 1e-6);    
end
```
This test fails on the first line. The `deserialize` function chokes on the referencedImageSeries element in the [header](https://github.com/ismrmrd/ismrmrd/blob/master/schema/ismrmrd_example_extended.xml). `deserialize` get's some length logic wrong, causing the error that ends the deserialization, but there is no easy fix. The operation attempted is at best is concatenation of the strings involved, not their storage as elements in an array. 

Likewise, this header simply breaks `deserialize`. 
```xml
<?xml version="1.0" encoding="utf-8"?>
<ismrmrdHeader xmlns="http://www.ismrm.org/ISMRMRD" xmlns:xs="http://www.w3.org/2001/XMLSchema"
               xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
               xsi:schemaLocation="http://www.ismrm.org/ISMRMRD ismrmrd.xsd">
   <experimentalConditions>
      <H1resonanceFrequency_Hz>63620740</H1resonanceFrequency_Hz>
   </experimentalConditions>
   <encoding>
      <encodedSpace>
         <matrixSize>
            <x>500</x>
            <y>500</y>
            <z>1</z>
         </matrixSize>
         <fieldOfView_mm>
            <x>500</x>
            <y>500</y>
            <z>1</z>
         </fieldOfView_mm>
      </encodedSpace>
      <reconSpace>
         <matrixSize>
            <x>500</x>
            <y>500</y>
            <z>1</z>
         </matrixSize>
         <fieldOfView_mm>
            <x>500</x>
            <y>500</y>
            <z>1</z>
         </fieldOfView_mm>
      </reconSpace>
[...]

Issue #152, opened 2020-05-01
ismrmrd-matlab serialize can not handle multiple reference scan labels


rajramasawmy:
My MRD has two reference scans: 'SenMap', and 'Noise'.

Just running a simple deserialze/serialize:
```matlab
dsetin = ismrmrd.Dataset(ProblemFile, 'dataset');

hdr = ismrmrd.xml.deserialize(dsetin.readxml);

xmlstring = ismrmrd.xml.serialize(hdr);
```
Throws this error:

```matlab
Error using ischar
Too many input arguments.

Error in ismrmrd.xml.serialize>append_node (line 299)

    if ischar(subheader.(name))

Error in ismrmrd.xml.serialize (line 55)

     append_node(docNode,node,dep,'dependencyType');
```

hansenms:
@rajramasawmy is this the same issue as #182? If so, can we close this one?

hansenms:
Closing per discussion since it is captured in #181

PR #200, opened 2022-03-16
Use cell arrays for multiple valued strings and dates in MATLAB deserialization of MRD header

kspaceKelvin:
The MATLAB representation of the MRD header is changed to use cell arrays when there are multiple values of strings and dates.  When there are single values, it is still a char string as before.  This addresses issue #152 and part of #181.  As @kristofferknudsen notes, this is a breaking change -- however this is only the case when there are multiple values, which wasn't properly supported before anyways.

Also fixed a few other fields that weren't properly serialized/deserialized.

Thoughts, @hansenms, @kristofferknudsen, @rajramasawmy ?


rajramasawmy:
Fine with me, we'll see when it breaks!

hansenms (approved):

PR #204, opened 2022-06-02
Add ImageHeader flag defines for first/last images of each loop counter

kspaceKelvin:
These defines are copied from the corresponding flag defines for the AcquisitionHeader.  These can be helpful for triggering processing of real-time image streaming, similar to how flags are used for raw data streaming.


dchansen:
So, we don't publish the packages on github (because there is no support for that), but we're ensuring that the code is packageable.
This means that [copr ](https://copr.fedorainfracloud.org/) can automatically pull, build and [publish](https://copr.fedorainfracloud.org/coprs/davidchansen/ISMRMRD/) packages for us.  This is the entrypoint for having ISMRMRD in Fedora, RedHat and derivatives. I'm also happy to set up a more official MRD account on copr that we could advertise more broadly.

Ideally we would have the same thing for Debian packages, but the launchpad build system is ducttapeandstringshopeyoureholdingitright kind of mess which is tricky to reproduce. I haven't given up on this yet though. 


As for the bug, it's unrelated to the pull request. Apparently an update to the checkout action broke our pipeline. I'm trying to find a workaround, but I can't guarantee I'll make it today. I'm ok with force-merging this PR though.

kspaceKelvin:
Sounds reasonable to keep the rpkg check for copr support, but as it's unrelated to this PR, I'll force merge and we'll troubleshoot for the next PR. 😄

hansenms:
Not sure I agree with the force merge here. We should never do that if it can be avoided.

PR #207, opened 2022-06-09
Added support for user encodinglimits

dchansen:
The encoding counters on acquisitions support up to 8 user counters, in addition to the regular ones like slice, phase and repetition.
These, however, do not have any corresponding encoding limits in the XML header.

This pull request rectifies that.


include/ismrmrd/xml.h
```diff
@@ -319,6 +320,7 @@ namespace ISMRMRD
     Optional<Limit> repetition;
     Optional<Limit> set;
     Optional<Limit> segment;
+    std::array<Optional<Limit>,ISMRMRD_USER_INTS> user;
```
hansenms, line 323: There is something a bit odd about this being an array here. I understand that they are all sort of the same, but you lay it out in the schema not as an array but individual variables names `user_1`, `user_2`, etc. and now we are taking the last string character and interpreting it as an array index.

dchansen: Sure. This is absolutely a compromise solution.
In the acquisition idx, the user variables are stored as an array, which is why I chose to mirror that in the EncodingLimits. 

We're fairly limited in what kind of datastructures we can have in the xml, and an array of optional values is not one of them. 

If you prefer, we can go with 

` Optional<std::array<Limit, ISMRMRD_USER_INTS>> ` instead. Then people would need to define all the limits, even if only one is used, but the xsl schema becomes more straightforward.

kspaceKelvin: I think Michael was thinking something more along the lines of ``user_1``, ``user_2``, etc, similar to how there's ``kspace_encode_step_1`` and ``kspace_encode_step_2`` here.  I could see the benefit of that, but am not strongly opinionated.  I prefer the committed version as opposed to the one in the comment, as I find the shorter resulting xml to be easier to read.

schema/ismrmrd_example_extended.xml
```diff
@@ -233,6 +233,11 @@
         <maximum>0</maximum>
         <center>0</center>
       </set>
+      <user_1>
+         <minimum>0</minimum>
```
kspaceKelvin, line 237: A little annoyed at the extra space, but we really need a PR that fixes whitespaces for this entire repo

include/ismrmrd/xml.h
hansenms: Yes, what @kspaceKelvin said. I am not speaking of the XML (yes it has many limits) it is how we expose it in the C++ API. If we have kspace_encode_step_1 and kspace_encode_step_2 and not `std::array<Limit> kspace_encode_step` then I think it would be more consistent to have it in the same way.

hansenms (approved):

PR #209, opened 2022-06-13
Added support for multiband slices. Added tests for the ISMRMRD header

dchansen:


valeryozenne:
Hi David, 

I'm writing a bit of context regarding the multiband reconstruction (mostly a copy of my previous email). Regarding our last conversation, I forget to mention the interleaved pattern could be a useful information for a generic reconstruction. Don't hesitate to ping me if additional information is needed. 

Regarding the naming, I honestly don't know: maybe FOV shift or z-CAIPI shift or slice CAIPI shift or CAIPI FOV shift (https://cni.stanford.edu/wiki/MUX_EPI) is more appropriate. I can invite few experts to join the discussion.

## Siemens sequences

On Siemens, there are two main sequences

* the Siemens Product or WIP referred as Product
* the CMRR sequence (used for the connectome project and largely available) referred as CMRR

The difference between both sequence implementations is mainly in the reference scan or (single band). In the case of the CMRR, the reference scans include the blip caipi shift (or phase offset) for each slice while in the case of the product, the reference scans do not include the blip caipi shift . As a consequence, to simulate the aliasing pattern and estimate the slice-grappa kernel, you need to:

* for the CMRR sequence:  sum the kspace

*  for the Product sequence: apply a specific phase offset per slice and then sum the kspace 

in the last case (Product), the phase offset has two components:
* the relative phase shift : related to the blip caipi shift factor, a device (proprietary) parameter that depend of the number of slices and of the acceleration factor  
* the absolute phase shift : related to the distance between the slice location and the isocenter

please note that in both case, you need to undo the phase offset once you have estimated the unaliased multiband images. Again the phase offset is only the relative phase shift for the CMRR and the  relative + absolute phase shift for the product. 

## Generic requirements

Thus, thinking of a generic solution, to do the multiband reconstruct, here are the parameters you need :

- the interleaved pattern 
- the multiband acceleration factor
- the blip caipi shift factor 
- the distance of the slice from the isocenter (only for the Product)

Here is some additional comment

*  the multiband acceleration factor is accessible through some IceParameter.  

* the interleaved pattern defining which slices are acquired together is a hard coded rule (that depends on the multiband acceleration factor and the number of slices). This information was not available to my knowledge, I did some reverse engineering to extract it but this is important information for a generic reconstruction. It could be stored as a vector in the MRD format. 

*  the blip caipi shift factor is accessible through some IceParameter . Please note that the blip caipi shift factor will depend of the in-plane acceleration, assuming 2 slices with GRAPPA=2 and Multiband=2 (MB=2), the usual caipi shift for MB=2 need to be multiplied by 2 to prevent the aliasing pattern from overlapping. Both version (CMRR or Product) differ in the definition of the caipi shift factor (with or without taking account the in-plane acceleration). 

* to compute the absolute phase shift, the distance of the slice from the isocenter is required for each slice. To compute it, you need the slice gap distance. To compute the slice gap distance , you need the slice thickness stored in the xml header and the slice location stored in acquisition header plus the knowledge of the interleaved pattern. This logic do not need to be reproduced. 

## Going further,

Please also consider that there is an extended version of MB called g-slicer (Martinos center). In this case, some dedicated RF pulses are used to encoded volumetrically the slab (to in fine increase the slice tickness resolution). It generate specific excitation profiles at a higher resolution that the slice tickness (https://pubmed.ncbi.nlm.nih.gov/28261904/). This means that each readout should be tagged to a specific excitation profiles. But I'm absolute not familiar with implementation.

Valéry

dchansen:
@kspaceKelvin Ready for final review

dchansen:
@kspaceKelvin Ping

hansenms (approved):

kspaceKelvin (commented):
Looks good!

PR #210, opened 2022-06-13
Documentation rework

kspaceKelvin:
Initial rework of documentation


hansenms (changes requested):
Overall this looks good. Some changes that I think should be made:

1. The rendering of all the data layout tables does not look good in rtd. We spoke about ways to add style, etc. to fix.
2. There are a number of links that don't work or render correctly. 
3. There is some redundancy in the MRD header description (under primitives) and the overview. I think that should be streamlined to have the information only in one place and have the other stuff reference.
4. The section on the use of the `MRD_MESSAGE_TEXT` should be removed IMO. I don't think we should be sending logs with the data at all.

kspaceKelvin:
@hansenms Hopefully I caught everything that we discussed today -- let me know if I missed anything.

hansenms:
@kspaceKelvin looks good overall.

One remaining thing is that for the "Message Types" section on the waveform, you still list the fields of the waveform header in a table. That is redundant since you have that table in the data primitives. Also the length of the waveform header says 240 bytes there, which should be 40, I believe. 

I would also say that I think the session protocol section is, unnecessarily, skewed towards socket communication, which is just one option as we discussed, but we can maybe evolve that as we go.

kspaceKelvin:
@hansenms Thanks for catching those two!  Fixed in the latest commit.  Agreed to rework the documentation towards non-socket communications in a future version.  Just need to play around with it a bit more myself first.

dchansen (approved):

PR #211, opened 2022-06-14
Add fixed PatientID and generated MeasurementID to header information in generate_cartesian_shepp_logan

Andrew-Dupuis:
The updated storage server fails to create or retrieve storage items if PatientID or MeasurementID aren't present in the header. This makes testing with the generate_cartesian_shepp_logan utility a bit harder. This PR adds a fixed PatientID of "ISMRMRDSheppLoganPhantom" and a random GUID MeasurementID to the header of the generated dataset. 

Before merging, I'd also like to clean up the whitespace issues, but I've left those alone for now to minimize # of lines changed and simplify review

### DIFF
--- a/.github/workflows/ismrmrd_windows_build.yml
+++ b/.github/workflows/ismrmrd_windows_build.yml
@@ -27,7 +27,7 @@ jobs:
     - name: Install dependencies
       shell: pwsh
       run: |
-        vcpkg install fftw3 boost-test boost-program-options boost-random hdf5
+        vcpkg install fftw3 boost-test boost-program-options boost-random boost-uuid hdf5
         vcpkg list
         dir c:\
         dir d:\
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,7 @@ endif()
 
 # compiler flags
 if (WIN32)
-    add_definitions(-DWIN32 -D_WIN32 -D_WINDOWS)
+    add_definitions(-DWIN32 -D_WIN32 -D_WINDOWS -DBOOST_UUID_FORCE_AUTO_LINK)
     add_definitions(-DUNICODE -D_UNICODE)
     add_definitions(-D_CRT_SECURE_NO_WARNINGS)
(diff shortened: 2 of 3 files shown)


hansenms (approved):

Andrew-Dupuis:
@hansenms any chance you could approve the workflow executions? Doesn't look like the integration set is running since I'm not part of this org.

Andrew-Dupuis:
Not totally sure why boost/uuid/uuid.hpp is missing on windows, but unless anyone has any ideas why that's popping up, will try to reproduce on a windows 10 machine when I have a spare moment

hansenms:
How about just passing in patientId and measurementId as command line parameters. I am not even sure it is completely predictable what will happen if we start generating these always. That way people can add them if they want.

Andrew-Dupuis:
@hansenms I tried adding the patient and measurement IDs via command line parameters, which worked - but reading the header back in failed because some of the required fields within measurementInfo and patientInfo would then be missing (which I didn't catch with the initial approach either). 

I just pushed an updated version to this branch that instead adds a "template" argument, that accepts an ISMRMRD XML header file as a template to which the generated phantom data (and header info) can be added. If the template argument isn't present, the behavior is identical to before. However, if a template is specified, the template is imported as an ISMRMRD header object, its encoding fields are removed so the phantom data can be appended, and any fields that are set later in the phantom generator are overwritten.

This allows you to use manually-specified headers where the missing fields are "populated" by the generator, or to use real headers from other datasets so you can test things like scanner/patient/date/site-specifics etc more easily. 

This seems like a better path than needing to keep specifying more command line parameters, but my implementation might not be ideal - very open to suggestions!

Andrew-Dupuis:
Example template, and resulting output header, below:
[template_xml.pdf](https://github.com/ismrmrd/ismrmrd/files/8920324/template_xml.pdf)
[output_xml.pdf](https://github.com/ismrmrd/ismrmrd/files/8920325/output_xml.pdf)

PR #214, opened 2022-08-06
Conda build

hansenms:
This PR introduces building conda packages into the build pipeline. So far, the only ismrmrd conda packages have been built by the Gadgetron project ([gadgetro-conda-recipes](https://github.com/gadgetron/gadgetron-conda-recipes), but it makes more sense to build the packages to allow us greater flexibility in the dependency targets and speed up the delivery of these packages. 

There is a new `conda/` folder which holds the scripts, etc. needed to build the packages. In this folder we introduce a small smoke test which generates a shepp logan phantom, reconstructs it and checks the results. 

There was a bug in the phantom generator introduced in #211. It did not add `patientPosition` to the `measurementInformation` field. Patient position is (apparently) a mandatory field. Not totally sure why and this should maybe be fixed in another PR.


.github/workflows/ismrmrd_conda.yml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Push conda package
+        shell: bash -l {0}
+        if: ${{ github.event_name == 'push' }}
+        env:
+          ANACONDA_TOKEN: ${{ secrets.ANACONDA_TOKEN }}
+        working-directory: conda
+        run: |
+          ./publish_package.sh -u ismrmrd -t "$ANACONDA_TOKEN" -p `find conda/build_pkg -name ismrmrd*.tar.bz2`
+
+  build-macos-conda-packages:
```
johnstairs, line 35: Did you consider using a ["matrix strategy"](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) instead?

roopchansinghv: This should actually be feasible here.

The only reason I separated out the macOS build into its own job, and did not use the build matrix, was that when I did that originally, I was not able to figure out how to filter the script and execution portion of the build configuration, by platform.  So if I used the matrix build strategy in the conda-recipes package, it would also then failed, b/c gadgetron would fail to build on macOS.

But since we now have parity on both platforms' builds, we should be able to use the built in matrix strategy, and just have 1 section/job for building the conda packages now for both platforms in parallel.

PR #218, opened 2022-09-07
always include stdint.h on Windows Visual Studio

### LINKED ISSUE #217: compilation problem on Windows VS 2022 and ninja: size_t
I'm getting a compilation problem as `size_t` is not defined.
```
C:\Users\krisf\Documents\devel\buildninja\SIRF-SuperBuild\sources\ISMRMRD\include\ismrmrd/ismrmrd.h(129): error C2061: syntax error: identifier 'ismrmrd_sizeof_data_type'
```
and many others. Weirdly, this doesn't happen when using the Visual Studio generator.

Currently there are the following lines
https://github.com/ismrmrd/ismrmrd/blob/bae6bcde22969d56e027de7497804520619b687c/include/ismrmrd/ismrmrd.h#L24-L36
I have no idea who is supposed to define `HAS_INT_TYPE`, but it isn't defined if I used either ninja or VS. So it uses these explicit defines, but they don't define `size_t`. It appears that when using the VS Generator, `size_t` is defined somehow, but not when using ninja.

These lines were introduced in the very first commit 10 years ago 8cd0ccdd36a47584bc5f912399fa09bc7589cb35, and then guarded with `HAS_INT_TYPE` in 8006a95cb23aae90430d82733d3a2ebad29f47fc, This looks very much outdated. I checked that everything works fine when I just do
```c++
 #include <stdint.h> 
 #include <stddef.h>     /* for size_t */ 
```
I don't know how old a compiler you want to support of course.

KrisThielemans:
remove obsolete code that tries to guess some typedefs for ints.

Fixes #217


dchansen (approved):
Looks good to me. Would like a comment from @hansenms or @kspaceKelvin to make sure we're not breaking anything important.

hansenms:
I am OK with this if @kspaceKelvin is. I do think we should rev the patch version (at least).

hansenms:
@kspaceKelvin what did we decide on this? Would this be OK for you?

hansenms:
@dchansen did you merge this into your other PR, so we can close this?

dchansen:
Yes. This has been merged.

PR #221, opened 2022-10-11
Serialization/Deserialization

hansenms:
This PR adds serialization and deserialization functions for the core data primitives (Acquisition, Image, Waveform). Additionally, it adds `ProtocolSerializer` and `ProtocolDeserializer` which are a set of classes that will produce a binary stream (including message ids) which would be consumed by an ISMRMRD/MRD application. 

To illustrate the use of those classes, this PR adds:
* `ismrmrd_hdf5_to_stream`: A command line tool, which will take an HDF5 file and convert it to a stream of data. It can stream raw data or images.
* `ismrmrd_stream_to_hdf5`: A command line tool, which consumes a stream and stores it as an HDF5 file. The stream can contain raw data and/or images.
* `ismrmrd_stream_recon_cartesian_2d`: A command line reconstruction tool, which mimics the other C++ reconstruction example, but it illustrates how to consume the data from stdin and produce images to stdout. 

Use these tools like this:

```bash
# Generate data
ismrmrd_generate_cartesian_shepp_logan

# Reconstruct
ismrmrd_hdf5_to_stream -i testdata.h5 | ismrmrd_stream_recon_cartesian_2d | ismrmrd_stream_to_hdf5 -o images.h5
```

With these tools it is now possible to orchestrate simple ISMRMRD/MRD reconstruction programs on the command line and remotely [as previously discussed](https://hackmd.io/@SUrbHI90Sp-pNc_b5_dhyw/By2h7r7Mo).

This PR also adds a devcontainer for VS code to make it easier to work with the repo.


libsrc/serialization.cpp
```diff
@@ -0,0 +1,214 @@
...
+    if (peek() != ISMRMRD_MESSAGE_WAVEFORM) {
+        throw std::runtime_error("Expected ISMRMRD_MESSAGE_WAVEFORM");
+    }
+    ISMRMRD::deserialize(wfm, _rs);
+    _peeked = ISMRMRD_MESSAGE_UNPEEKED;
+}
+
+} // namespace ISMRMRD
+
+// template instanciations
```
johnstairs, line 182:
```suggestion
// template instantiations
```

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,83 @@
...
+      "settings": {
+        "files.trimFinalNewlines": true,
+        "files.trimTrailingWhitespace": true,
+        "files.insertFinalNewline": true,
+        "go.toolsManagement.checkForUpdates": "local",
+        "go.useLanguageServer": true,
+        "gopls": { "ui.semanticTokens": true },
+
+        "[cpp]": {
+          "editor.formatOnSave": true
```
johnstairs, line 26: Not sure this is a good idea for this codebase.

include/ismrmrd/serialization.h
```diff
@@ -0,0 +1,161 @@
...
+
+// A wrapper, which we can use wrap any object that has a read(...) method
+class ReadableStream {
+public:
+    void read(char *buffer, size_t count) {
+        reader_impl->read(buffer, count);
+    }
+
+    template <class HAS_READ>
+    ReadableStream(HAS_READ &has_read) {
```
johnstairs, line 43: Did you consider making these `shared_ptr`s intead of references?

utilities/ismrmrd_hdf5_to_stream.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    try {
+        std::string xml;
+        d.readHeader(xml);
+        ISMRMRD::IsmrmrdHeader hdr;
+        ISMRMRD::deserialize(xml.c_str(), hdr);
+        serializer.serialize(hdr);
+    } catch (std::exception &e) {
+        if (image_series.empty()) {
+            // rethrow exception if image series not specified
```
johnstairs, line 54: Why are we otherwise swallowing the exception?

include/ismrmrd/serialization.h
```diff
@@ -0,0 +1,161 @@
...
+        writer_impl->write(buffer, count);
+    }
+
+    template <class HAS_WRITE>
+    WritableStream(HAS_WRITE &has_write) {
+        // make_unique is C++14
+        writer_impl = std::unique_ptr<WritableStreamT<HAS_WRITE> >(new WritableStreamT<HAS_WRITE>(has_write));
+    }
+
+    virtual ~WritableStream() {}
```
johnstairs, line 81: Are these needed?

include/ismrmrd/serialization.h
johnstairs: You could have both:
```c++
  template <class HAS_WRITE>
  WritableStream(std::shared_ptr<HAS_WRITE> has_write)
      : writer_impl_(new WritableStreamT<HAS_WRITE>(has_write)) {
  }

  template <class HAS_WRITE>
  WritableStream(HAS_WRITE& has_write)
      : writer_impl_(new WritableStreamT<HAS_WRITE>(std::shared_ptr<HAS_WRITE>(&has_write, [](HAS_WRITE*) {}))) {
  }
```

dchansen: shared_ptr would not work with C++98. 
It would also make [ownership somewhat unclear](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f7-for-general-use-take-t-or-t-arguments-rather-than-smart-pointers) - but maybe we should make it clear that ReadableStream is a view? ReadableStreamView maybe?

Either way, in general, a reference is all we can ask for. Many I/O resources are not copyable and sometimes not even movable. 

My thought was actually that we would always rely on the implicit conversion to ReadableStream, and just call the functions with the raw types. I know there is an overhead in creating the unique_ptr inside, but it should be very small.

johnstairs: If we want to target C++98 then you can forget everything I said. 

My original thinking was to prevent someone from writing code like this (which will result in bad things):
``` cpp
std::unique_ptr<WritableStream> CreateWritableStream() {
    std::ofstream ofs("foo.bin", std::ios::binary);
    return std::make_unique<WritableStream>(ofs);
}
```

By instead requiring them to do something like
``` cpp
std::unique_ptr<WritableStream> CreateWritableStream2() {
    auto ofs = std::make_shared<std::ofstream>("foo.bin", std::ios::binary);
    return std::make_unique<WritableStream>(ofs);
}
```

But of course this a shared ptr can also result in bad things when it attempts to destruct:
``` cpp
std::unique_ptr<WritableStream> CreateWritableStream3() {
    auto ofs = std::shared_ptr<std::ostream>(&std::cout);
    return std::make_unique<WritableStream>(ofs);
}
```
And the proper way to wrap cout in a shared_ptr is pretty cumbersome:
``` cpp
std::unique_ptr<WritableStream> CreateWritableStream3() {
    auto ofs = std::shared_ptr<std::ostream>(&std::cout, [](std::ostream*){});
    return std::make_unique<WritableStream>(ofs);
}
```

I like your WritableStreamView suggestion. I think that helps make the lack of ownership clear.

I also think being able to rely on implicit conversion seems like a neat idea.

libsrc/serialization.cpp
```diff
@@ -0,0 +1,214 @@
+#include <sstream>
+#include <string>
+
+#include "ismrmrd/serialization.h"
+
+namespace ISMRMRD {
+
+void serialize(const Acquisition &acq, WritableStream &ws) {
+    AcquisitionHeader ahead = acq.getHead();
+    ws.write(reinterpret_cast<const char *>(&ahead), sizeof(AcquisitionHeader));
```
johnstairs, line 10: I might be paranoid here, but I would add static assertions that classes such as `AcquisitionHeader` that are writing out in this way have the exact layout that we are expecting. `is_standard_layout`, the overall `sizeof` and the `offsetof` each member variable.

libsrc/serialization.cpp
```diff
@@ -0,0 +1,214 @@
...
+    uint64_t attr_length = img.getAttributeStringLength();
+    ws.write(reinterpret_cast<char *>(&attr_length), sizeof(uint64_t));
+    if (attr_length) {
+        ws.write(img.getAttributeString(), ihead.attribute_string_len);
+    }
+    ws.write(reinterpret_cast<const char *>(img.getDataPtr()), img.getDataSize());
+}
+
+void serialize(const Waveform &wfm, WritableStream &ws) {
+    ws.write(reinterpret_cast<const char *>(&wfm.head), sizeof(ISMRMRD_WaveformHeader));
```
johnstairs, line 31: More paranoia: I'd want to be sure that `head` is actually an `ISMRMRD_WaveformHeader`

include/ismrmrd/serialization.h
hansenms: Quick question here. If we are truly targeting C++98, we probably have to get rid of the unique_ptr too and just have a `ReadableStreamView*` member that we manage ourselves?

libsrc/serialization.cpp
```diff
@@ -0,0 +1,214 @@
...
+    ws.write(reinterpret_cast<const char *>(wfm.begin_data()), wfm.head.number_of_samples * wfm.head.channels * sizeof(uint32_t));
+}
+
+void deserialize(Acquisition &acq, ReadableStream &rs) {
```
johnstairs, line 35: The deserialize functions are not taking into account a premature EOF, and silently return invalid data.

include/ismrmrd/serialization.h
dchansen: You're right. It is rather annoying to have to do that ourselves. 
We should perhaps discuss where our API boundary should go. After all, we're not implementing this feature for the C API, so we could make the streaming interface here C++11 only. @kspaceKelvin any thoughts?

What we **need** is to be C++98 compatible for the basic types and the xml-header. I'm working on how we might do that without breaking API compatibillity.

include/ismrmrd/serialization.h
```diff
@@ -0,0 +1,179 @@
...
+ * The standalone function that serialize to and deserialize from a stream will
+ * serrialize the data structures without adding "message id" identifiers.
+ * The ProtocolSerializer and ProtocolDeserializer classes are used to to create streams
+ * that include the message id in front of each message.
+ *
+ */
+
+namespace ISMRMRD {
+
+const uint16_t ISMRMRD_MESSAGE_UNPEEKED = 0;
```
dchansen, line 28: Would this be better as an enum? (Very much not an enum class)

libsrc/serialization.cpp
dchansen: I think it would be nice, but should probably happen in the main Ismrmrd header? After all, it is a pretty important guarantee.

dchansen (commented):
I generally think it looks pretty good - Just don't like the stdin thing

utilities/ismrmrd_hdf5_to_stream.cpp
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    if (vm.count("help")) {
+        std::cerr << desc << "\n";
+        return 1;
+    }
+
+    ISMRMRD::Dataset d(input_file.c_str(), groupname.c_str(), false);
+
+#ifdef _MSC_VER
```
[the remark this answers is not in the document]
hansenms: Yeah, lemme noodle on this. I think what I will do is to let the stream to/from hdf5 tools take files in and out and optionally (explicit opt in with, e.g. `--use-stdout`) use stdin/out so those tools can operate like that. For the recon example, I will adapt to run and have the e2e tests use files on all platforms and probably an e2e tests validating that piping works on linux/mac. I am exploring the windows situation a bit more and will come up with something.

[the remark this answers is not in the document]
dchansen: Sounds good to me

include/ismrmrd/ismrmrd.h
```diff
@@ -132,89 +131,88 @@ EXPORTISMRMRD size_t ismrmrd_sizeof_data_type(int data_type);
  * Acquisition Flags
  */
 enum ISMRMRD_AcquisitionFlags {
-    ISMRMRD_ACQ_FIRST_IN_ENCODE_STEP1               =  1,
```
johnstairs, line 135: Was all the whitespace removal intentional?

libsrc/serialization.cpp
hansenms: I went through and added static_asserts in the `ismrmrd.h` and `waveform.h` headers. So we now have the checks. Unfortunately, I also made the discovery that the waveform header has an 8-byte boundary while the other structures use 2-byte boundaries. So that header is 40 bytes intead of 32 and the version field (which is a 16 byte field) actually takes up 8 bytes. One can debate which boundary is the best, I think it is a problem that it is inconsistent between the different headers. This 8-byte boundary is baked in in many places, so we can't change it now, but the documentation that @kspaceKelvin is working on does not reflect this discrepancy, so it should be updated to reflect this.

hansenms: @dchansen was it accidental that it ended up like that or?

include/ismrmrd/ismrmrd.h
hansenms: Nope, that as was a mistake. Switching off automatic formatting for these files.

.devcontainer/devcontainer.json
hansenms: Removed.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,83 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
+// https://github.com/microsoft/vscode-dev-containerdevcotns/tree/v0.238.0/containers/go
+{
+  "name": "ismrmrd",
+  "build": {
+    "dockerfile": "../Dockerfile",
```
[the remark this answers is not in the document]
hansenms: Yep, makes sense.

include/ismrmrd/serialization.h
hansenms: I have made this C++98 compatible for now. It is one point to manage in this case.

libsrc/serialization.cpp
dchansen: Yes - absolutely an accident. I didn't stop to consider the layout implications of the design when it was originally made. By the time I realised, it was already released and in use.

hansenms: We are checking now that the current layout doesn't change. Also added standard layout checks on the C++ classes that inherit from the base structs. All of that is wrapped in C++11 checks.

libsrc/serialization.cpp
hansenms: Yep, added a check for that.

include/ismrmrd/serialization.h
hansenms: Removed.

utilities/ismrmrd_hdf5_to_stream.cpp
hansenms: The try is on whether or not we can get a header and write it. If we are not streaming images, we have to have a header, so we will rethrow whatever happened. If we do have images, it is OK, we can continue.

libsrc/serialization.cpp
hansenms: I have added an `eof()` on the readable view interface and `bad()` on the writable stream view. They correspond to the relevant functions on iostream. You would have to implement those if you wrap something else.

libsrc/serialization.cpp
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// Helper function that deserializes attributes and pixels
+template <typename T>
+void deserialize_attr_and_pixels(Image<T> &img, ReadableStream &rs) {
+    uint64_t attr_length;
+    rs.read(reinterpret_cast<char *>(&attr_length), sizeof(uint64_t));
+    if (attr_length) {
+        char *attr = new char[attr_length + 1];
+        rs.read(attr, attr_length);
```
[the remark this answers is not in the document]
hansenms: Done

hansenms:
OK @johnstairs and @dchansen, I *think* I have addressed the issues:

1. I have added layout checks of the structs.
2. I have added eof and bad checks on the (de)serialization.
3. I have added file interfaces (in addition to stdin/stdout interfaces on the added utilities).
4. The end to test tests check both file and stdin/stdout workflows on all 3 target platforms. On Windows I am now running the tests from a cmd shell since PowerShell does crazy UTF-16 things with stdin/out.

Lemme know if you have other concerns/comments. I think it is pretty close now.

libsrc/serialization.cpp
```diff
@@ -0,0 +1,239 @@
...
+}
+
+ProtocolDeserializer::ProtocolDeserializer(ReadableStreamView &rs) : _rs(rs), _peeked(ISMRMRD_MESSAGE_UNPEEKED) {}
+
+uint16_t ProtocolDeserializer::peek() {
+    if (_peeked == ISMRMRD_MESSAGE_UNPEEKED) {
+        _rs.read(reinterpret_cast<char *>(&_peeked), sizeof(uint16_t));
+        if (_peeked == ISMRMRD_MESSAGE_IMAGE) {
+            _rs.read(reinterpret_cast<char *>(&_peeked_image_header), sizeof(ImageHeader));
+        }
```
johnstairs, line 142: I think we should check eof here

utilities/ismrmrd_stream_to_hdf5.cpp
```diff
@@ -0,0 +1,148 @@
...
+    if (vm.count("input")) {
+        std::ifstream is(input_file.c_str(), std::ios::binary);
+        if (!is) {
+            std::cerr << "Error: Could not open input file " << input_file << std::endl;
+            return 1;
+        }
+        convert_stream_to_hdf5(output_file, groupname, is);
+    } else if (use_stdin) {
+        ISMRMRD::set_binary_io();
+        convert_stream_to_hdf5(output_file, groupname, std::cin);
```
johnstairs, line 141: I think we should make sure we have consumed the entire stream at this point.

hansenms: Good suggestion. Guess the question is what is the behavior if there is more data in the stream? It would happen if there we hit close before the end. Should we start over and keep consuming? Not sure that makes sense, so an error?

johnstairs: Safest is probably to raise an error.

libsrc/serialization.cpp
hansenms: Done

utilities/ismrmrd_stream_to_hdf5.cpp
hansenms: Done

utilities/ismrmrd_stream_to_hdf5.cpp
```diff
@@ -88,6 +79,11 @@ void convert_stream_to_hdf5(std::string output_file, std::string groupname, std:
             throw std::runtime_error(ss.str());
         }
     }
+
+    // If we can read any more at this point, it is an error
+    if (is.rdbuf()->in_avail() > 0) {
```
johnstairs, line 84: I could be wrong, but isn't this a function of the internal buffering state, not necessarily the true end of stream?

johnstairs (approved):

dchansen:
I'm somewhere sunny and far away from a proper computer. Can I do the final review Monday?

hansenms:
@dchansen, of course!

hansenms:
Per discussions in the meeting today, I have added/fixed a few things:

1. I have simplified the `ReadableStreamView` and `WritableStreamView` stuff so that it is just an interface and we are providing implementations of that interface for iostream. That way the path with things like boost asio should be simpler and it doesn't complicate things here.
2. I have added serialization/deserialization of config file (e.g. default.xml), config text (aka config script), and text message (used by some systems for logging).
3. I have appropriate functions for those to the protocol serializer/deserializer classses. 
4. Made some tweaks to the HDF5 to stream adapter, so that you can now pass a `-c` (config file) or `-C` (local config) in (optionally) and if you do, they will serialized to the stream. That allows the hdf5/stream adapter tools to be a complete Gadgetron client, for example, if you have a Gadgetron running on 9004, you can do (effectively assemble a client in the shell):

    ```bash
    ./utilities/ismrmrd_hdf5_to_stream -i testdata.h5 -c default.xml --use-stdout | socat - TCP4:localhost:9004 | ./utilities/ismrmrd_stream_to_hdf5 images.h5 --use-stdin -o images.h5
    ```
5. The stream recon will log a config file if we receive it, but we don't do anything with it otherwise, it is ignored.
6. The stream to hdf5 converter will log (to stderr) any text messages it receives. They have no representation in the HDF5 file.

@dchansen have a look and lemme know what you think.
@kspaceKelvin this should help (maybe a lot) with your concerns.
@johnstairs please have another look.

libsrc/serialization.cpp
```diff
@@ -42,6 +42,22 @@ void serialize(const Waveform &wfm, WritableStreamView &ws) {
...
 
+void serialize(const char (&str)[1024], WritableStreamView &ws) {
+    ws.write(str, 1024);
+    if (ws.bad()) {
+        throw std::runtime_error("Error writing fixed length char array to stream");
+    }
+}
+
+void serialize(const std::string &str, WritableStreamView &ws) {
+    uint32_t len = str.length();
```
dchansen: I hate to say this, but in order to be compatible with Gadgetron et al, you actually need to send the null terminator. 
str.length()+1 here.

include/ismrmrd/serialization.h
```diff
@@ -70,11 +79,34 @@ EXPORTISMRMRD void deserialize(Image<T> &img, ReadableStreamView &rs);
 // deserialize Waveform from istream
 EXPORTISMRMRD void deserialize(Waveform &wfm, ReadableStreamView &rs);
 
+// deserialize const length (1024) char array from istream. Used for CONFIG FILE
+EXPORTISMRMRD void deserialize(char (&str)[1024], ReadableStreamView &rs);
```
dchansen, line 83: I think this should explicitly take a "ConfigFile" object, rather than a char array. They are pretty hard to work with in practice, as they decay to char* as soon as they can. 
Alternatively, I would use std::array<char,1024>.

dchansen (changes requested):
Looks good to me - just a few minor tweaks and it's good to go.

libsrc/serialization.cpp
hansenms: Hmmm....I don't think so. If I look at a couple of places, e.g.:

String in config script:
https://github.com/gadgetron/gadgetron/blob/master/apps/clients/gadgetron_ismrmrd_client/gadgetron_ismrmrd_client.cpp#L1248

String in text message:
https://github.com/gadgetron/gadgetron/blob/master/apps/gadgetron/connection/Writers.cpp#L22

I don't see an extra null terminator and this code does actually work with the text messages that Gadgetron will log back and with config script.

include/ismrmrd/serialization.h
hansenms: Done. I went with ConfigFile in the interface, `std::array` is C++11, which is why I had it as a fixed length char.

hansenms:
@dchansen can you check your notes on the string serialization (used for config script and text message). I have gotten this wrong more than once, so I appreciate the consult.

libsrc/serialization.cpp
dchansen: Hmm.... ok, I can see it's no longer there (used to be though). 
[The documentation does say null terminated, however.](https://ismrmrd--210.org.readthedocs.build/en/210/mrd_messages.html#mrd-message-config-text), and I remember having to fight with python to remove the extra null.
I think ICEGadgetron might even rely on the strings being correctly null terminated, but I could be misremembering.

I've wanted to get rid of that null terminator since I saw it, so if we actually can, I say let's do it! 
@kspaceKelvin does this work for you? The null terminator is a buffer overflow waiting to happen.

hansenms: I just chatted with @kspaceKelvin, he can chime in. I doubt that IceGadgetron or IceFIRE rely on the null terminator (since the Gadgetron is, in fact, not sending it). Quick look at python code seems to indicate that it is reads the length specified and then splits on null terminator (which may or may not be there) so it works either way. I think we should get rid of it now that we have the chance. I will let @kspaceKelvin respond too. But my choice would be to take the PR version.

kspaceKelvin: I wrote the null-terminator in the documentation I also seem to recall it being used in Gadgetron when I started with the Python/MATLAB stuff.  I agree with both of you, @dchansen, @hansenms -- in the documentation and reference code, let's take out the null-terminator during serialization for [MRD_MESSAGE_CONFIG_TEXT](https://ismrmrd--210.org.readthedocs.build/en/210/mrd_messages.html#mrd-message-config-text), [MRD_MESSAGE_HEADER](https://ismrmrd--210.org.readthedocs.build/en/210/mrd_messages.html#mrd-message-header), [MRD_MESSAGE_TEXT](https://ismrmrd--210.org.readthedocs.build/en/210/mrd_messages.html#mrd-message-text), and the MetaAttributes in [MRD_MESSAGE_IMAGE](https://ismrmrd--210.org.readthedocs.build/en/210/mrd_messages.html#mrd-message-image).  I believe it has to stay for [MRD_MESSAGE_CONFIG_FILE](https://ismrmrd--210.org.readthedocs.build/en/210/mrd_messages.html#mrd-message-config-file), as there's no other way to determine when the file name ends in the 1024 character array.

We should make sure that the deserializers don't break if there is a null-terminator in any of the above though.

kspaceKelvin:
> Per discussions in the meeting today, I have added/fixed a few things:
> 
> 1. I have simplified the `ReadableStreamView` and `WritableStreamView` stuff so that it is just an interface and we are providing implementations of that interface for iostream. That way the path with things like boost asio should be simpler and it doesn't complicate things here.
> 2. I have added serialization/deserialization of config file (e.g. default.xml), config text (aka config script), and text message (used by some systems for logging).
> 3. I have appropriate functions for those to the protocol serializer/deserializer classses.
> 4. Made some tweaks to the HDF5 to stream adapter, so that you can now pass a `-c` (config file) or `-C` (local config) in (optionally) and if you do, they will serialized to the stream. That allows the hdf5/stream adapter tools to be a complete Gadgetron client, for example, if you have a Gadgetron running on 9004, you can do (effectively assemble a client in the shell):
>    ```shell
>    ./utilities/ismrmrd_hdf5_to_stream -i testdata.h5 -c default.xml --use-stdout | socat - TCP4:localhost:9004 | ./utilities/ismrmrd_stream_to_hdf5 images.h5 --use-stdin -o images.h5
>    ```
> 5. The stream recon will log a config file if we receive it, but we don't do anything with it otherwise, it is ignored.
> 6. The stream to hdf5 converter will log (to stderr) any text messages it receives. They have no representation in the HDF5 file.
> 
> @dchansen have a look and lemme know what you think. @kspaceKelvin this should help (maybe a lot) with your concerns. @johnstairs please have another look.

This does indeed help a lot with my concerns and I like the symmetry of being about to use the proposed pipes as a transparent alternative to network sockets.  Your points regarding separating out the control plane are well taken and I think we should continue that discussion outside of this PR.

libsrc/serialization.cpp
hansenms: Good call @kspaceKelvin. I added a test to make sure that if we get the string `hello\0` (length 6), we end up with `hello` (length 5) when deserializing. It actually revealed that we didn't handle that perfectly, so I fixed the string deserializer.  

@dchansen let me know if you have other concerns. Otherwise, I just need a sign off from you and I will get this merged.

dchansen (approved):

PR #223, opened 2022-11-13
Buffered reading and writing MRD files, resulting in 6x speedup on read and 4x on write

dchansen:
This PR makes HDF5 reuse the internal buffers used for reading from a dataset.
This reduces the time it takes to read individual acquisitions by 6x (at least on Windows) and makes reading individual acquisitions as fast as reading the entire dataset in one go.

The cost is that datasets now carry 2Kb of buffer state, which I think is acceptable.


dchansen:
On linux, the improvement is more modest - on the order of 30%.

tests/test_dataset.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+float create_random_float() {
+    return dist(rng);
+}
+
+BOOST_AUTO_TEST_SUITE(DatasetTest)
+
+
+BOOST_AUTO_TEST_CASE(test_acquisition_header) {
+
```
hansenms, line 22: nit: formatting and additional line breaks in this file.

hansenms (changes requested):
Look OK in general. I mainly have some questions on whether this should be optional and what the right size of that buffer is. Maybe if those things were configurable, it would be easier to test with the benchmark app? Also some thoughts on handling versioning with the adde vcpkg configuration.

vcpkg.json
```diff
@@ -0,0 +1,18 @@
+{
+  "$schema": "https://raw.githubusercontent.com/microsoft/vcpkg/master/scripts/vcpkg.schema.json",
+  "name": "ismrmrd",
+  "version": "1.12.0",
```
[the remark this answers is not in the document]
dchansen: Managed to get CMake to actually check that this matches.

tests/benchmark_dataset.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+static boost::random::mt19937 rng;
+static boost::random::uniform_real_distribution<float> dist = boost::random::uniform_real_distribution<float>();
+
+float create_random_float() {
+    return dist(rng);
+}
+
+int main(int argc, char** argv) {
+
```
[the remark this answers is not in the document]
dchansen: Done

dchansen:
@hansenms This should be ready to merge I think, unless you have any other comments?

libsrc/dataset.c
```diff
@@ -152,6 +152,11 @@ static int delete_var(const ISMRMRD_Dataset *dset, const char *var) {
     return status;
 }
 
+#define ISMRMRD_READ_BUFFER_SIZE 1024*1024
+
+
+static char ismrmrd_conversion_buffer[ISMRMRD_READ_BUFFER_SIZE];
```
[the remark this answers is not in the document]
dchansen: Yes. This is baked into HDF5. The "multithreaded" version is just using a global lock.

libsrc/dataset.c
```diff
@@ -152,6 +152,11 @@ static int delete_var(const ISMRMRD_Dataset *dset, const char *var) {
     return status;
 }
 
+#define ISMRMRD_READ_BUFFER_SIZE 1024*1024
```
johnstairs, line 155: How did you arrive at this number?

tests/benchmark_dataset.cpp
```diff
@@ -0,0 +1,52 @@
...
+    for (size_t i = 0; i < acqs.size(); i++) {
+        std::generate((float *)acqs[i].data_begin(), (float *)acqs[i].data_end(), create_random_float);
+        std::generate((float *)acqs[i].traj_begin(), (float *)acqs[i].traj_end(), create_random_float);
+    }
+
+    {
+
+        auto start = std::chrono::high_resolution_clock::now();
+        Dataset dataset = Dataset(temp.string().c_str(), "/test", true);
+        for (size_t i = 0; i < acqs.size(); i++)
```
hansenms, line 33: nit: Added `{}` to the for loop.

tests/test_dataset.cpp
```diff
@@ -0,0 +1,56 @@
...
+static boost::random::uniform_real_distribution<float> dist = boost::random::uniform_real_distribution<float>();
+
+float create_random_float() {
+    return dist(rng);
+}
+
+
+BOOST_AUTO_TEST_SUITE(DatasetTest)
+
+BOOST_AUTO_TEST_CASE(test_acquisition_header) {
```
hansenms, line 20: Think that test name is copied from the acquisition tests. Maybe choose a better name.

hansenms (approved):
Some minor additional comments, but otherwise looks good.

libsrc/dataset.c
dchansen: It's the same as the default value in HDF5. 
I haven't played around with it too much, but 1024 seemed to give some degradation in write performance on linux.

PR #225, opened 2023-01-20
Fixed a bug where the diffusion header was not correctly read

dchansen:


dchansen:
So... I'm guessing the MacOS builder was updated? 
I don't really have access to a Mac, so my abillity to debug this is limited, but maybe it is time to say goodbye to our own packaged pugixml?

roopchansinghv:
This (i.e. GitHub's macOS builders being updated) is most likely the case.  I've seen the above error in Gadgetron's macOS build locally (i.e. on my own Macs) with newer compilers.

This particular error is not that bad - Clang is complaining about sprintf being deprecated for snprintf, and this seems to have been remedied in later versions of pugixml.  But this (presence of 'sprintf') is also a more significant issue in Gadgetron, so we need to figure out what to do there as well.  Gadgetron's macOS build isn't yet fully working, but this might also be one of the things causing the remaining issues there ...

hansenms:
@dchansen yes, I think we should remove (or at least update) the pugi that we bundle. It is a bit of a security problem to have since we haven't updated in a long time. Also, there may be a license issue?

hansenms (approved):

PR #226, opened 2023-01-20
Remove PugiXML sources from ISMRMRD and configure build to use system library.

roopchansinghv:
The majority of changes are to the environment configuration and build setup.

There are a few instances of functions in newer versions of PugiXML being deprecated, and having to be replaced with newer version, namely original 'load' method having to be replaced by 'load_string'.


roopchansinghv:
@hansenms and @dchansen - so this all seems to work with building across all platforms.  Let me know if this looks okay, or if you would want anything other than a patch version increment to finalize this PR?

@hansenms - I tried to leverage the "USE_SYSTEM_PUGIXML" in the top-level CMakeLists.txt file - and I was able to patch the conda build pipelines, as well as the base CMake and Windows builds to take advantage of this.  However, I could not figure out how to make the RPM package build use this option.  So I backed out these changes, and entirely removed the option to use a locally supplied version of PugiXML itself.  Along with all of this was some minor code updates, to replace 'load' (deprecated in the PugiXML library itself) with 'load_string'.

@dchansen - I think you should now be able to cleanly overlay your fixes for diffusion, on top of these changes.

dchansen (approved):

PR #229, opened 2023-02-09
Updating build docs to be consistent with removal of Pugixml sources from ISMRMRD

roopchansinghv:
This also aligns ISMRMRD's docs with the automated package building pipelines.


hansenms:
We should add pugixml to the apt install line in the same file.

hansenms (approved):

PR #237, opened 2023-06-13
Bumping build number - PugiXML conda break

hansenms:
This PR bumps the build number of the conda package. 

The conda-feedstock package for PugiXML was actually a static library, which we depended on, but they decided to move to a shared library, but they did it without bumping the version number of the package. This will pull in this change and create a new build version for us. It is to reduce the downstream breakage in various packages where build strings are not pinned. 

See: https://github.com/conda-forge/pugixml-feedstock/pull/22

There will be a subsequent package to pull in a later version of PugiXML and rev the version of ISMRMRD.

### DIFF
--- a/include/ismrmrd/xml.h
+++ b/include/ismrmrd/xml.h
@@ -6,17 +6,18 @@
 #ifndef ISMRMRDXML_H
 #define ISMRMRDXML_H
 
-#include "ismrmrd/export.h"
 #include "ismrmrd.h"
+#include "ismrmrd/export.h"
+#include <array>
 #include <cstddef>
-#include <new> //For std::badalloc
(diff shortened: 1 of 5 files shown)


johnstairs (approved):

include/ismrmrd/xml.h
```diff
@@ -31,164 +32,157 @@
  * @{
  */
 
-namespace ISMRMRD
-{
+namespace ISMRMRD {
```
johnstairs, line 35: It looks like this file was reformatted. Maybe do that in a separate PR?

hansenms: Yes, reverted that.

PR #238, opened 2023-06-13
Rev pugixml version, switch to mamba in devcontainer

hansenms:
This PR ups the version for PugiXML and revs the package version for the library.

We also switch to mamba in the devcontainer.


johnstairs:
We'll want to set the build number back to 0 after merging #237

johnstairs (approved):

PR #242, opened 2023-09-06
C++98 compatibility

hansenms:
This PR returns C++98 compatibility to the majority of the code. We do not validate this on all platforms, but we run a smoketest to see that it builds on Linux with `-stdc++98`. There would likely be some work to ensure that this works on a given platform. 

The main changes are:
1. Move semantics is encapsulated with a C++11 check.
2. Enum class are encapsulated with a C++11 check and an drop in alternative is provided. 
3. There is a header file with some shimming in case we are not C++11.
4. Smoketest pipeline for C++98.
5. Minor tweaks to loops, etc. 
6. Removing `auto`.
7. Bump version.

The majority of changes is in C++ parsing code.


dchansen (approved):
Looks good to me, except for the namespace std bit.

hansenms:
@dchansen, finally got around to addressing your comments. Have another look. 

This version now compiles across a range of Siemens IDEA versions.

libsrc/xml.cpp
```diff
@@ -790,430 +807,443 @@ void append_optional_three_dimensional_float(pugi::xml_node& n, const char* chil
...
+bool operator>=(const MeasurementDependency &lhs, const MeasurementDependency &rhs) {
+    return !(lhs < rhs);
+}
+bool operator==(const ReferencedImageSequence &lhs, const ReferencedImageSequence &rhs) {
+    return lhs.referencedSOPInstanceUID == rhs.referencedSOPInstanceUID;
+}
+bool operator!=(const ReferencedImageSequence &lhs, const ReferencedImageSequence &rhs) {
+    return !(rhs == lhs);
+}
+bool operator==(const MeasurementInformation &lhs, const MeasurementInformation &rhs) {
```
naegelejd, line 1098: This line and similar below are very long, hard to read. The ones above (e.g. `StudyInformation`) are much cleaner.

naegelejd (approved):
Looks all good

Issue #244, opened 2023-10-09
Conda recipe is missing boost dependency version


naegelejd:
The ISMRMRD conda package does not specify which version of Boost is needed at runtime. It should specify the version that was used to build the package.

Steps to reproduce the behavior, e.g.:
1. Install `ismrmrd` using `conda` _without_ explicitly installing a specific version of `boost`.
2. Invoke `ismrmrd_generate_cartesian_shepp_logan` in a shell
3. You will likely see `... error while loading shared libraries: libboost_program_options.so.1.80.0: cannot open shared object file: No such file or directory` because your system probably has a more recent version of Boost installed (e.g. 1.82.0).

**System Information:**
 - OS: Ubuntu 20.04
 - `conda` environment that specifies an `ismrmrd` (e.g. version 1.13.7) dependency but does NOT specify a boost dependency

**Additional context**
You can manually override this issue by explicitly specifying a dependency on `boost=1.80.0` in your conda `environment.yml`, but we can fix this properly here in the ISMRMRD conda packaging setup.

PR #245, opened 2023-10-09
Specify Boost runtime version in conda-build recipe

naegelejd:
Simple fix for #244 . See https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html#general-pinning-examples.


hansenms:
I think we want to bump the patch version here, otherwise will not actually push a new package.

hansenms (approved):

hansenms:
The readthedocs thing is failing. Probably same issue as https://github.com/gadgetron/gadgetron/pull/1237/files

PR #250, opened 2024-04-19
Added fovShifted field to the header

dchansen:


dchansen:
On the call, there was quite a bit of discussion about why this flag was a bool, rather than a value defining how much the FOV had been shifted.
@cmirkes I think perhaps you could better explain the motivation than I was able to?

cmirkes:
Let me provide some background information: To achieve field-of-view (FOV) shifts, a linearly increasing phase needs to be added to the k-space raw data. The phase increment can be added during acquisition by the scanner hardware or software before storing the data to disk. For Cartesian imaging, almost all vendors apply the FOV shift before saving the raw data. As a matter of fact, the FOV shift does not need to be applied during image reconstruction. In case of EPI, the FOV shift might be applied for the phase-encoding direction [kspace_encode_step_1] but not for the read direction [kspace_encode_step_0] (e.g. because ramp sampling is used). For spiral imaging, none of the vendors applies a FOV shift and the data needs to be modulated during image reconstruction to achieve the correct FOV shift. The amount the data needs to be shifted by is already contained in the acquisition header of each acquisition line. Hence, the only missing information is whether the scanner has already manipulated the data or not. This information cannot be extracted from any current parameter of the ISMRMRD XML header.

PR #253, opened 2024-07-05
add serialization for ndarray

xueh2:
add serialization and test for ndarray
increase version to 1.14.1


libsrc/serialization.cpp
```diff
@@ -58,6 +58,33 @@ void serialize(const std::string &str, WritableStreamView &ws) {
     }
 }
 
+template <typename T> 
+void serialize(const NDArray<T> &arr, WritableStreamView &ws) {
+    uint16_t ver = arr.getVersion();
```
hansenms, line 63: What is the purpose of this version?

hansenms:
I think this has to be added to the protocol serializer and deserializer too. And the ismrmrd_hdf5_to_stream and ismrmrd_stream_to_hdf5 have to support it as well, this raises the question of whether these arrays need some kind of identifier that would put them in separate "datasets" in the HDF file and what the ordering of all of this should be. Same problem as we deal with for the waveforms with timestamps, etc.

libsrc/serialization.cpp
xueh2: it is in the NDArray definition of ismrmrd.

xueh2:
Please review and merge.

xueh2:
also, how to update conda ismrmrd channel to install this version (1.14.1)?

xueh2:
@hansenms could we move forward with this PR? Need it for data loader for ai training.

hansenms (approved):

PR #256, opened 2024-11-22
Fix locale serialization issue

### LINKED ISSUE #255: Serialization output differs depending on locale
Different user locales can result in different serialization output, both of the XML Header and MetaContainer.

In particular, the `en_US.UTF-8` locale causes integers to be serialized with commas separating every three digits. The underlying serialization uses `std::stringstream`, which depends on locale settings.

This bug only surfaces when a client program using `libismrmrd` changes the global locale. The ISMRMRD utilities do *not* do this, so they don't exhibit the bug. Gadgetron, however, does do this, so setting env `LANG=en_US.UTF-8` then running Gadgetron will result in ISMRMRD datasets that may not be compatible outside of that locale.

You *could* reproduce this within this repo by setting the user-defined locale in an ISMRMRD utility, such as `generate_cartesian_shepp_logan.cpp`:
```cpp
int main(int argc, char** argv)
{
    std::locale::global(std::locale(""));
    ...
}
```
Then compare the output with different settings for `LANG`:
```bash
export LANG=C
rm -f testdata.h5 && ./utilities/ismrmrd_generate_cartesian_shepp_logan && h5dump -d /dataset/xml testdata.h5 | grep H1
Generating Cartesian Shepp Logan Phantom!!!
Acceleration: 1
                        <H1resonanceFrequency_Hz>63500000</H1resonanceFrequency_Hz>

export LANG=en_US.UTF-8
rm -f testdata.h5 && ./utilities/ismrmrd_generate_cartesian_shepp_logan && h5dump -d /dataset/xml testdata.h5 | grep H1
Generating Cartesian Shepp Logan Phantom!!!
Acceleration: 1
                        <H1resonanceFrequency_Hz>63,500,000</H1resonanceFrequency_Hz>
```

naegelejd:
Added a unit test and also tested within Gadgetron, where #255 was first discovered.

Closes #255

### DIFF
--- a/include/ismrmrd/meta.h
+++ b/include/ismrmrd/meta.h
@@ -129,6 +129,7 @@ namespace ISMRMRD {
             l_ = l;
             d_ = static_cast<double>(l_);
             std::stringstream strstream;
+            strstream.imbue(std::locale::classic());
             strstream << l_;
             strstream >> s_;
         }
@@ -137,6 +138,7 @@ namespace ISMRMRD {
             d_ = d;
             l_ = static_cast<long>(d_);
             std::stringstream strstream;
+            strstream.imbue(std::locale::classic());
             strstream << d_;
             strstream >> s_;
         }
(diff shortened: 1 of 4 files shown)


hansenms (approved):
Bump version?

